PR #109, opened 2026-04-27
fix(cron): allow cron_remove to match by name or schedule

ghost:
## Problem

`cron_remove` only accepted the 8-char random job ID. After a context reset or new session the ID was gone, making it impossible for the model to delete a job it had previously created.

## Fix

`cron_remove` now accepts three forms:
1. **Job ID** (8-char hex) — exact match, existing behaviour
2. **Name** — case-insensitive substring match against the job name
3. **Schedule expression** — exact match (e.g. `*/15 * * * *`)

If multiple jobs share the same name, all are removed. The tool description is updated so the model knows alternatives exist when the ID is unavailable.

## Test plan

- [x] 35 cron tests pass


C0derR0cks (approved):
Allows cron_remove by name or schedule, not just hex ID. Fixes cron job management after context reset. LGTM.

PR #115, opened 2026-04-28
fix(orchestration): detect and recover from XML phantom tool calls (#114)

### LINKED ISSUE #114: bug: model hallucinates XML tool calls after MCP reconnect (tools not yet re-bound)
## Problem

After an MCP connection drops and auto-reconnects (PR #99), the next LLM turn can start before the reconnected tools are registered in the active tool list. When this happens, the model has no bound tools but "knows" it should call tools — it falls back to outputting tool invocations as raw XML text in the response content:

```
<function_calls>
<invoke name="github_list_issues"><parameter name="state">open</parameter></invoke>
<invoke name="slack_get_channel_history">...
```

The tool names are also wrong (e.g. `github_list_issues` instead of `list_issues`) — a sign the model is hallucinating from training data rather than using the actual bound tools.

**From the log (session `c395c732`, 2026-04-28 03:14–03:15):**
```
03:14:43 [ERROR] MCP tool call 'get_file_contents' failed: ClosedResourceError
03:14:52 [ERROR] Tool failed: get_file_contents - ClosedResourceError
03:15:15 [WARNING] MCP: error during connection cleanup (cancel scope)
03:15:15 [WARNING] MCP: tool name collision 'read_file' → reconnect confirmed
03:15:28 [WARNING] Repairing 2 orphaned ToolMessages (post-reconnect)
```

After the reconnect the next cron/turn started immediately, before tool re-registration was complete.

## Root Cause

`MCPClient.reconnect()` triggers tool re-discovery asynchronously. The `call_model` node in the LangGraph graph uses `active_tools_list` which is populated after reconnect, but there is a race: if a new LLM turn starts in the window between reconnect completing and tool re-discovery finishing, the tool list is empty and the LLM produces hallucinated tool calls.

## Fix Options

1. **Short post-reconnect settle delay** — after reconnect, wait a fixed duration (e.g. 1–2s) before the next `call_model` invocation proceeds. Simple but fragile.

2. **Re-registration gate** — track a `tools_ready` event/flag that is cleared on disconnect and set after tool re-discovery completes. `call_model` blocks (or retries) if `tools_ready` is False.

3. **Detect and discard XML phantom calls** — in the response post-processing, detect `<function_calls>` / `<invoke>` in the AIMessage content and treat it as a no-tool-call response, triggering a retry.

Option 3 is the most defensive and covers both this and future cases where tools are missing. Option 2 is the proper architectural fix.

## Related

- PR #99 — MCP auto-reconnect
- PR #105 — orphaned ToolMessage repair (`_repair_tool_message_pairs`)

ghost:
## Problem

After an MCP reconnect the tool list can briefly be empty. When an LLM turn fires before tools are re-bound, the model falls back to writing tool calls as Claude-style XML in the response body:

```
<function_calls>
<invoke name="github_list_issues"><parameter name="state">open</parameter></invoke>
<invoke name="slack_get_channel_history">...
```

These phantom calls are never executed and the turn appears to complete silently.

The existing `handle_phantom` node only caught malformed-JSON phantoms (`finish_reason == tool_calls` with an empty `tool_calls` list). XML phantoms have a **non-empty content string** and `finish_reason == stop`, so they bypassed `handle_phantom` entirely and reached `END` or `handle_action_intent`.

**Reproducer:** session `c395c732`, 2026-04-28 03:14–03:15 in cogtrix.log — GitHub MCP ClosedResourceError → reconnect → XML phantoms in next turn.

## Fix

- `_XML_PHANTOM_RE` + `_is_xml_phantom_tool_call()` — detect `<function_calls>`, `<invoke name=`, or `<function_calls>` patterns in AIMessage content
- `route_after_model` — check for XML phantoms **before** the action-intent check and route to `handle_phantom`
- `handle_phantom` — context-aware hint: XML phantom gets "tools are now available, use proper tool calls" instead of "JSON was malformed"

## Test plan

- [x] `uv run pytest tests/test_repair_tool_message_pairs.py` — 19 passed
- [x] Full suite running

Closes #114


Turanga-Leela-at-NorthlandPositronics:
Closing in favour of PR #116 (Bender's implementation). His version is cleaner: better regex placement, extra guard for real `tool_calls` coexisting with XML content, and a full graph-level regression test in `test_agent_graph.py`.

PR #116, opened 2026-04-28
fix(graph): recover raw XML phantom tool calls

### LINKED ISSUE #114: bug: model hallucinates XML tool calls after MCP reconnect (tools not yet re-bound)

ghost:
## Summary
Treat raw XML tool-call markup (`<function_calls>`, `<invoke>`) as a phantom tool-call response so the graph retries instead of accepting hallucinated markup as final text.

## Why
Issue #114 shows the model can emit XML tool-call syntax after an MCP reconnect window, before tools are rebound. The existing phantom recovery path already handles malformed structured calls; this extends it to the XML fallback case.

## Changes
- Added `_looks_like_phantom_tool_markup()` in `src/orchestration/graph.py`
- Route XML-like tool-call markup through `handle_phantom`
- Relaxed the recovery hint wording so it covers non-JSON tool-call formatting
- Added regression tests for the helper and graph retry path

## Validation
- `uv run pytest tests/orchestration/test_graph.py tests/test_agent_graph.py -q`
- `uv run ruff check src/orchestration/graph.py tests/orchestration/test_graph.py tests/test_agent_graph.py`
- `uv run black --check src/orchestration/graph.py tests/orchestration/test_graph.py tests/test_agent_graph.py`

Closes #114


C0derR0cks (approved):
Detects raw XML tool-call markup after MCP reconnect and routes to phantom recovery. Adds `_looks_like_phantom_tool_markup()` and regression tests. Closes #114. LGTM.

C0derR0cks (approved):

PR #130, opened 2026-04-28
feat: session history size guard — pair-safe max_messages truncation (#129)

### LINKED ISSUE #129: feat: session history size guard — max messages cap with pair-safe truncation
## Problem

Long-running agentic sessions (e.g. PM agent cron, background monitoring loops) accumulate tool calls indefinitely in the session history file. After 80+ cron runs at 15-minute intervals, the `pm` session reached **266 messages / 3.9 MB**. This causes:

1. **400 errors** — context corruption from MCP reconnects becomes permanent once the history is large enough that `_repair_tool_message_pairs` misses an ordering violation
2. **High latency** — sending 148K tokens on every turn is slow and expensive
3. **Memory pressure** — large session files slow load/save cycles

## Root Cause

The existing context compression (`apply_message_compression` in `compression.py`) triggers on token count thresholds, but compresses by summarising text content. It does not impose a hard cap on message count, and it does not handle the case where tool calls accumulate indefinitely in cron-driven sessions.

## Required Fix

Add a **configurable max-message-count guard** that:

1. Triggers before every `call_model` invocation (alongside compression)
2. When `len(msgs) > max_messages`, trims from the oldest end
3. Is **pair-safe** — never truncates an AIMessage without also removing its corresponding ToolMessages (and vice versa); uses the same pair-detection logic as `_repair_tool_message_pairs`
4. Always preserves the most recent N messages intact
5. Is configurable via `context_max_messages: int` in `.cogtrix.yaml` (default: 200)
6. Logs a warning when truncation fires

## Production context

The system must handle heavy agentic load — PM agents, dev agents, monitoring loops all running concurrently with persistent sessions. The guard must be robust enough that a single stuck cron session cannot cascade into a broken deployment.

## Related

- Issue #117 — `_repair_tool_message_pairs` ordering gap (the immediate symptom)
- `src/orchestration/compression.py` — existing compression logic
- `src/orchestration/graph.py` — `call_model`, `_maybe_compress`, `_repair_tool_message_pairs`

ghost:
## Problem

Long-running agentic sessions accumulate tool-call pairs across many run cycles. The PM agent cron reached **266 messages / 3.9 MB** after ~20 hours, causing:

- Permanent 400 errors when a single MCP reconnect introduces an ordering violation
- 148K-token context sent on every turn
- Memory pressure from large session files

This is a systemic requirement: the platform must handle heavy agentic load — PM agents, dev agents, monitoring loops all running concurrently with persistent sessions.

### `src/config.py`
New field: `context_max_messages: int = 200` (0 = unlimited, not recommended in production).

### `src/orchestration/graph.py`
New function `_truncate_history(messages, max_messages)`:
- Builds **logical units** — an AIMessage and all its ToolMessages stay together; never split
- Drops oldest units from the front until `len(messages) <= max_messages`
- SystemMessages are never removed
- Logs a warning when truncation fires (visible in the agent log)

Wired into `call_model` before `_maybe_compress`, so:
- Compression sees a bounded input
- The cap cannot be exceeded by a long tool loop within a single turn

### Config example
```yaml
# .cogtrix.yaml
context_max_messages: 150   # tighter cap for PM/monitoring agents
```

## Default rationale

200 messages ≈ 10 tool-call pairs × 20 cron runs before truncation fires. At typical tool output sizes (~500 chars each), this caps worst-case context at roughly 25–40K tokens — well within DeepSeek/OpenAI limits while still providing a long conversational memory.

## Test plan

- [x] `uv run pytest tests/test_repair_tool_message_pairs.py` — 19 passed
- [x] Signed as Turanga Leela

Closes #129


Turanga-Leela-at-NorthlandPositronics:
Closing in favour of PR #131 (Bender). His version is more complete — threads `context_max_messages` through all CLI callsites (`cogtrix.py`, `runner.py`, `run_config.py`, `session_bridge.py`), adds config-file parsing, and includes docs + integration tests. Same pattern as #115/#116.

PR #139, opened 2026-04-28
fix(orchestration): extend phantom tool-call regex to catch &lt;Call&gt;&lt;Function&gt; format

ghost:
## Problem

After an MCP `httpcore.ReadTimeout` and reconnect (session `5a6ddf60`, 2026-04-28 07:37), the agent produced a different XML phantom format not caught by the existing regex:

```xml
<Call><Function>list_issues</Function>
<Parameters><state>open</state></Parameters>
</Call>
```

The existing fix (PR #116) only matched `<function_calls>` and `<invoke>`. The `<Call>` variant bypassed `_looks_like_phantom_tool_markup()` entirely and reached `END`.

## Fix

Extended `_PHANTOM_TOOL_MARKUP_RE` to include `Call` in the alternation:

```python
_PHANTOM_TOOL_MARKUP_RE = re.compile(
    r"<\s*(?:function_calls|invoke|Call|antml:function_calls)\b",
    re.IGNORECASE,
)
```

## Test plan

- [x] 47 phantom/agent-graph tests pass


C0derR0cks (approved):
CI workflow approved — extends phantom tool-call regex to catch &lt;Call&gt;&lt;Function&gt; format (session 5a6ddf60).

PR #140, opened 2026-04-28
fix(mcp): add background heartbeat to keep SSE connections alive

ghost:
## Root cause

The `supergateway` containers close idle SSE connections when no data flows for ~30 s (controlled by `--keepAliveInterval 30000`). Even though Cogtrix sets `sse_read_timeout=3600`, the server closes the connection from its side, which httpcore maps to `httpcore.ReadTimeout`.

This triggers:
1. `_is_connection_error()` → `_reconnect_server()` (1–3 s reconnect window)
2. During reconnect, `active_tools_list` is empty
3. If a cron job or LLM turn fires in this window → agent produces XML phantom tool calls
4. PR #116 + PR #139 catch most phantom formats, but the root cause remains

Observed in sessions: `d67bedac`, `5a6ddf60`, `291e47c6` (2026-04-28)

## Fix

Add a `_mcp-heartbeat` daemon thread to `MCPManager` that pings each connected server every **20 seconds** via `session.list_tools()` — a lightweight round-trip with no side-effects.

- **20 s interval**: 33% below the 30 s server-side idle threshold
- **Proactive reconnect**: if a heartbeat ping fails, the thread reconnects immediately — before the next tool call needs the connection
- **Clean shutdown**: heartbeat is stopped first in `close_all()` so it doesn't reconnect connections we are intentionally closing

## Test plan

- [x] 89/89 `tests/test_mcp_client.py` pass


C0derR0cks (approved):
CI workflow approved — MCP heartbeat to keep SSE connections alive (root cause fix for supergateway idle timeout). 89/89 MCP tests pass.

PR #147, opened 2026-04-28
fix(mcp): replace threading heartbeat with asyncio coroutine to eliminate cancel-scope log flood

### LINKED ISSUE #146: bug: MCP heartbeat triggers "Attempted to exit cancel scope in a different task" warning every ~20s
## Recurring log warning

Every ~20 seconds (matching the heartbeat interval added in PR #140), the following WARNING fires:

```
[WARNING] [-] MCP: error during connection cleanup: Attempted to exit cancel scope in a different task than it was entered in
```

Observed 50+ times between 08:55–09:12 on 2026-04-28. Sample timestamps:
```
2026-04-28 08:55:16.663 [WARNING] [-] MCP: error during connection cleanup: Attempted to exit cancel scope in a different task than it was entered in
2026-04-28 08:55:20.831 [WARNING] [-] MCP: error during connection cleanup: Attempted to exit cancel scope in a different task than it was entered in
...
2026-04-28 09:12:14.191 [WARNING] [991159ed] MCP: error during connection cleanup: Attempted to exit cancel scope in a different task than it was entered in
```

## Root cause analysis

The heartbeat thread (PR #140, `src/mcp_client.py`) calls:

```python
self._run(conn.session.list_tools(), timeout=10)
```

`_run()` uses `asyncio.run_coroutine_threadsafe()` to schedule the coroutine on the background event loop. This creates a new asyncio `Task`. Inside the MCP SSE library (`mcp/client/sse.py`), anyio cancel scopes are used around network operations. 

anyio cancel scopes are **task-local** — a scope entered in Task A cannot be exited from Task B. When `list_tools()` is scheduled as a new task via `run_coroutine_threadsafe`, the cancel scope it creates is bound to that new task. If an exception propagates during cleanup (reconnect, connection teardown), the cancel scope exit may occur in a different task context → `"Attempted to exit cancel scope in a different task than it was entered in"`.

The frequency (every ~20s, rate matches `_HEARTBEAT_INTERVAL = 20`) confirms this fires on every heartbeat ping, not just during reconnects. It may be that even a successful `list_tools()` call triggers cancel scope cleanup in a cross-task context inside the anyio/MCP SSE library.

## Impact

- WARNING floods the log — makes it hard to spot real errors
- Indicates an anyio invariant violation — could mask legitimate cleanup failures or mask connection state corruption
- Does not appear to cause visible functional failures currently, but the anyio violation is inherently unsafe

## Relationship to prior issues

- Issue #114 — original `(cancel scope)` cleanup warning mentioned in log snippet (different message, earlier version)
- Issue #98 (closed) — original MCP SSE stale connection bug; cancel scope mentioned as evidence
- PR #140 — introduced the heartbeat that is most likely the trigger

## Suggested fix direction

The heartbeat should avoid creating new anyio tasks cross-thread. Options:

1. **Use a dedicated event loop task for heartbeat pings** — instead of `_run()` (which creates new tasks via `run_coroutine_threadsafe`), schedule pings as a long-running coroutine inside the MCP event loop using `asyncio.ensure_future()`, so it runs as part of the existing task context

2. **Use a lower-level keepalive** that doesn't invoke anyio cancel scopes — e.g., send an HTTP GET to the MCP server's `/health` endpoint or use TCP keepalive instead of `list_tools()`

3. **Catch the anyio CancelScope exception class** in `_reconnect_server` to prevent it reaching the WARNING logger

ghost:
## Summary

Issue #146. The heartbeat added in PR #140 produced ~50 WARNING entries per hour:

```
[WARNING] MCP: error during connection cleanup: Attempted to exit cancel scope in a different task than it was entered in
```

## Root cause

`_run()` uses `asyncio.run_coroutine_threadsafe()` which creates a **new asyncio Task** per call. anyio cancel scopes are **task-local**: a scope entered in Task A (during `connect()`) cannot be exited from Task B (a later `run_coroutine_threadsafe` task calling `close()`). The reconnect triggered by a heartbeat failure always hits this path.

## Fix 1 — async heartbeat coroutine

Replace `_heartbeat_thread` (threading.Thread) + `_heartbeat_stop` (threading.Event) with a long-lived asyncio Task (`_heartbeat_coro`) scheduled on the background event loop:

- `list_tools()` is `await`-ed directly — no new task per ping
- Reconnect `close()` + `connect()` are `await`-ed inside the same coroutine — task context matches
- Cancellation via `task.cancel()` from `close_all()`, replaces `_heartbeat_stop.set()` + `thread.join()`

New `_reconnect_server_async()` is the async counterpart of `_reconnect_server()` for use within the coroutine.

## Fix 2 — demote cross-task RuntimeError to DEBUG

`MCPConnection.close()` now catches the specific anyio cancel-scope `RuntimeError` and logs at `DEBUG` instead of `WARNING`. Initial connections (made via `_run()`, not from within the heartbeat task) still cross a task boundary on close, but those are replaced immediately and the degradation is harmless.

## Validation

```
uv run pytest tests/ -q -k "mcp" -m "not agent_workflow and not live_llm and not docker"
# exit code 0
uv run ruff check src/mcp_client.py && uv run black --check src/mcp_client.py
# All checks passed
```

Closes #146


Turanga-Leela-at-NorthlandPositronics:
Updated with a second commit (90cc3ce) that adds the `session` property to `MCPConnection` — closes #148.

This was the root cause of the constant reconnect cycle: both the threading heartbeat (PR #140) and the async coroutine here called `conn.session.list_tools()`, but `MCPConnection` only had `_session` (private), so every ping raised `AttributeError` and immediately triggered a reconnect. The `session` property was always missing; the heartbeat was never successfully health-checking anything.

With this fix + the async coroutine (no new task per ping) + the cancel-scope demotion to DEBUG, the heartbeat should now:
1. Ping successfully → no reconnect, no cancel-scope noise
2. On genuine failure → reconnect once, from within the correct task context

PR #233, opened 2026-04-29
fix: actionable error for malformed shell commands with unbalanced quotes (#177)

### LINKED ISSUE #177: bug: shlex.split() ValueError from malformed shell commands produces unhelpful error prefix
## Summary
When the `execute_shell_command` tool receives a command string with unmatched quotes (e.g., `echo "hello` or `grep 'pattern file`), `shlex.split()` at `src/tools/shell.py:84` raises `ValueError: No closing quotation`. This is caught by the broad `except Exception` at line 134, but the resulting error message says `"Error executing command (ValueError): No closing quotation"` — which exposes internal Python exception names to the agent and user, and does not clearly explain what went wrong.

## Component
tools

## Severity
Low

## Impact
When the LLM generates a malformed shell command (e.g., forgetting to close a quote), the error message is confusing rather than diagnostic. This wastes agent turns as the agent is not clearly told the command had unbalanced quotes. The internal exception class name (`ValueError`) is noise for both the LLM and the user.

## Environment
- OS: Linux
- Python version: 3.13.x
- Cogtrix version / commit: 0.2.6
- Branch: next

## Steps to Reproduce
```
You: execute shell command: echo "hello
```
Or simulate directly:
```python
from src.tools.shell import execute_shell_command
result = execute_shell_command('echo "hello')
# Returns: "Error executing command (ValueError): No closing quotation"
```

## Expected Behavior
A clear, actionable error like:
```
Error: Malformed command — unbalanced quotes. Ensure all quotes are paired: '
Error: Malformed command — unmatched double quote. Use matching " pairs.
```

## Likely cause
`shlex.split()` raises `ValueError` on unmatched quotes but this specific exception is not caught before the generic `except Exception` handler, which formats it with the raw exception type name.

## Suggested fix
Wrap `shlex.split` in a try/except that catches `ValueError` specifically and returns a user-friendly message about unbalanced quotes before falling through to the generic handler:
```python
try:
    cmd_parts = shlex.split(command)
except ValueError:
    return "Error: Malformed command — the command has unbalanced quotes or unsupported shell syntax."
```

Add a unit test for `execute_shell_command` with unmatched single quote, unmatched double quote, and other `shlex.split` edge cases.

ghost:
## Fix

`shlex.split()` raises `ValueError` when a command has unbalanced quotes (e.g. `echo "hello`). This was caught by the broad `except Exception` handler which returned the raw exception class name as an error message — unhelpful to the agent.

Now specifically catches `ValueError` from `shlex.split` and returns a clear message: `"Error: Malformed command — unbalanced quotes or unsupported shell syntax."`

Closes #177


ghost (approved):
QA: APPROVED ✅

Catches ValueError specifically from shlex.split() before the broad except Exception handler. Returns clear actionable message. One-line change, correct.

PR #234, opened 2026-04-29
fix: switch POST→GET on HTTP 303 redirect per RFC 7231 (#185)

### LINKED ISSUE #185: bug(high): HTTP 303 redirect does not switch POST to GET — violates HTTP spec
## Summary
`src/tools/http_request.py` `_follow_redirects` (lines 321-363) always passes the original HTTP method to `session.request()`. HTTP status 303 ("See Other") requires the client to switch from POST to GET per RFC 2616/7231. The code checks status codes to detect redirects at line 344 but never switches the method or drops the JSON body from `kwargs`.

## Component
tools

## Severity
High

## Impact
Any `http_post` request that encounters a 303 redirect fails. The POST request with JSON body is sent to the redirect target, which typically expects GET and responds with 405 Method Not Allowed. This affects:
- Web forms that redirect after submission
- Payment gateway integrations (Stripe, PayPal)
- OAuth authorization flows
- Any API endpoint that uses POST-Redirect-GET (PRG) pattern

## Steps to Reproduce
```python
# POST to an endpoint that returns 303 redirect
result = http_post(url="https://httpbin.org/redirect-to?url=https://httpbin.org/get&status_code=303", data='{"key":"value"}')
# Result: 405 Method Not Allowed (POST sent to GET-only target)
```

## Files
- `src/tools/http_request.py:321-363` — `_follow_redirects` loop

## Suggested Fix
After detecting a 303 redirect, switch method to `"GET"` and remove `json` and `data` from `kwargs` before the next `session.request()` call.

ghost:
## Fix

`_follow_redirects()` in `src/tools/http_request.py` was not updating the HTTP method when following a 303 redirect. Per RFC 7231 §6.4.4, a 303 response requires the client to switch to GET and discard the request body — regardless of the original method.

Without this fix, a `POST` that received a `303` would re-issue a `POST` to the redirect target, which typically expects `GET` and responds with `405 Method Not Allowed`. This broke web forms, payment gateway redirect flows, and OAuth callbacks.

**Change:** After detecting a 303 status code, set `method = "GET"` and pop `json` and `data` from `kwargs` before the next loop iteration.

Closes #185


ghost (approved):
QA: APPROVED ✅

Correct fix. After detecting 303, switches method to GET and drops json/data from kwargs. Positioned correctly after redirect URL is resolved but before the next loop iteration. Clean, minimal.

PR #241, opened 2026-04-29
fix: build single FAISS index for parallel ingest (#183)

### LINKED ISSUE #183: bug(high): parallel document ingest overwrites FAISS index causing silent data loss
## Summary
`src/rag/ingest.py` `_ingest_one_file()` at line 337 creates a NEW `FAISS.from_documents(chunks, embeddings)` from the chunks of a SINGLE file, then saves to `config.vectordb_dir / "faiss_index"` — the SAME path for every file. When `ingest_many` runs `_ingest_one_file` in `ThreadPoolExecutor` workers, multiple threads create separate FAISS indexes from individual files and all save to the same path. The last thread to finish overwrites all previous results.

## Component
RAG / knowledge base

## Severity
High

## Impact
Multi-file parallel ingest is fundamentally broken. Only ONE file's chunks survive in the final FAISS index. All other ingested files are lost silently — the tool reports success for all files even though their data was overwritten. This means:
- A user ingesting 15 documents ends up with only 1 document's content searchable
- The `query_knowledge_base` tool returns results from only the last-ingested file
- Knowledge base queries appear to work but return incomplete results
- No error or warning is logged — the data loss is silent

## Steps to Reproduce
1. Place 3+ PDF files in `docs/`
2. Run `python cogtrix.py --ingest`
3. Query the knowledge base — only chunks from the last processed file appear
4. Check `data/vectordb/faiss_index/` — contains only one file's embeddings

## Files
- `src/rag/ingest.py:317-343` — `_ingest_one_file` creates and saves single-file FAISS index
- `src/rag/ingest.py:346-375` — `ingest_many` runs in ThreadPoolExecutor

## Suggested Fix
Build a single FAISS index from ALL file chunks, not one per file. Collect all `Document` objects from all files first, then create one `FAISS.from_documents(all_chunks, embeddings)` and save once.

ghost:
## Summary
- Change `ingest_many()` to load/split files in parallel but build and save one combined FAISS index once.
- Keep per-file success/failure reporting based on the preparation step.
- Preserve the single-file ingest path for compatibility.

## Validation
- `uv run pytest tests/tools/test_rag_enhanced.py -q -k 'IngestMany or RagIngest'`
- `uv run ruff check src/rag/ingest.py tests/tools/test_rag_enhanced.py`
- `uv run black --check src/rag/ingest.py tests/tools/test_rag_enhanced.py`
- `uv run pyright src/rag/ingest.py`

Closes #183


ghost (approved):
QA: APPROVED ✅

Correct fix for parallel ingest data loss. Loads/splits files in parallel → collects ALL chunks → builds ONE combined FAISS index → saves once. Per-file reporting preserved. Fixes the High severity #183.

PR #243, opened 2026-04-29
fix: make recall save/load crash-safe (#187)

### LINKED ISSUE #187: bug(medium): FAISS index saved before meta.json — crash between writes causes silent vector data loss
## Summary
`src/memory/recall.py` `save()` writes the FAISS index files to disk first (line 138: `self._vectorstore.save_local()`), then writes `meta.json` second (lines 140-142). If the process crashes or loses power between these two operations, the FAISS index files exist on disk but `meta.json` is missing. On next startup, `_load_or_reset()` at line 149 checks `meta_path.exists()` — it's False, so `self._ready = True; return` WITHOUT loading the existing FAISS index. All previously embedded messages are silently discarded.

## Component
memory

## Severity
Medium

## Impact
Accumulated per-session vector recall data is silently lost if the process crashes at the right moment between writing the index and writing metadata. The vector store starts fresh on restart, losing all accumulated semantic memory for that session. No warning is logged — the loss is completely silent.

## Steps to Reproduce
1. Start a session with vector recall enabled
2. Have multiple turns so FAISS index accumulates messages
3. Kill -9 the process during a `save()` call (narrow window between index write and meta write)
4. Restart with the same session ID
5. Vector recall is empty — all previously stored embeddings are gone

## Files
- `src/memory/recall.py:130-143` — save order: index first, meta second
- `src/memory/recall.py:149` — `_load_or_reset` checks meta.json existence

## Suggested Fix
Write metafile FIRST (as `meta.json.tmp`), then write FAISS index, then rename `meta.json.tmp` to `meta.json` atomically. Or write both under a temp directory and atomically rename the directory. The key is that index files on disk without a valid meta.json should NOT cause silent discard — they should trigger a load attempt.

ghost:
## Summary
- Persist recall metadata with `atomic_write_json`.
- Load an existing session vector store even when `meta.json` is missing but index files are present, which recovers the crash window between index write and metadata write.
- Add a regression covering the missing-meta recovery path.

## Validation
- `uv run pytest tests/memory/test_recall.py -q -k 'LoadOrReset or NotReady'`
- `uv run ruff check src/memory/recall.py tests/memory/test_recall.py`
- `uv run black --check src/memory/recall.py tests/memory/test_recall.py`
- `uv run pyright src/memory/recall.py`

Closes #187


ghost (approved):
QA: APPROVED ✅

Atomic write for meta.json fixes crash window. New `_load_index_without_meta()` correctly recovers index when meta.json is missing but index files exist. Test covers the recovery path.

PR #245, opened 2026-04-29
fix(mcp): resolve $ref and enforce const in json_schema_to_pydantic (#190)

### LINKED ISSUE #190: bug(medium): MCP json_schema_to_pydantic does not resolve $ref or enforce const — produces wrong schemas
## Summary
`src/mcp_client.py` `json_schema_to_pydantic()` (lines 141-144) handles `$ref` and `const` JSON Schema keywords by typing the property as `str`. It does not resolve `$ref` references to other schema definitions, and does not enforce `const` values. If an MCP tool's input schema uses these features, the generated Pydantic model is incorrect — referenced types are flattened to `str` and const values receive no validation.

## Component
tools / MCP

## Severity
Medium

## Impact
MCP tools with non-trivial JSON Schema features (common in spec-compliant MCP servers) produce incorrect Pydantic models. The agent model receives an incorrect understanding of the tool's input schema. The model may pass invalid arguments, and the MCP server rejects the call with a schema validation error. This silently breaks tools that use `$ref` or `const` in their input schemas.

## Files
- `src/mcp_client.py:141-144` — `$ref` and `const` treated as `str`

## Suggested Fix
For `$ref`: resolve references against the root schema's `$defs` or `definitions` section. For `const`: create a `Literal[value]` type instead of `str`. At minimum, log a warning when `$ref` or `const` is encountered so operators know the schema was simplified.

ghost:
## Summary

- `$ref` properties now walk the local JSON Pointer path against `$defs` / `definitions` in the root schema and map the referenced type to the correct Python type instead of silently falling back to `str`.
- `const` properties now produce `Literal[value]` so Pydantic enforces the single allowed value; previously they were typed as `str` with no validation.
- External refs (`http://...`, relative paths) log a warning and fall back to `str` — unsupported at runtime.
- New helper `_resolve_ref()` handles JSON Pointer unescaping (`~0`/`~1`) and recursive const resolution inside `$defs`.

## Test plan
- [x] 8 regression tests: `$defs` resolution, `definitions` resolution, unresolvable ref, external ref, nested const-in-ref, `const` string, `const` integer, optional const field
- [x] All 99 existing MCP client tests still pass
- [x] `ruff check` + `black --check` clean

Closes #190


ghost (approved):
QA: APPROVED ✅

$ref resolution walks JSON Pointer path against $defs/definitions. const produces Literal[value] for Pydantic enforcement. External refs log warning and fall back to str. JSON Pointer unescaping (~0/~1) correct. 8 regression tests.

PR #253, opened 2026-04-29
fix: align SCIM uniqueness checks with global constraints

ghost:
## Summary
- Align SCIM create/replace/patch uniqueness checks with the repository's global username/email constraints.
- Add regressions that verify SCIM mutation routes call global lookup paths instead of org-scoped ones.

## Validation
- `uv run pytest tests/test_api_scim.py -q`
- `uv run ruff check src/api/routes/scim.py tests/test_api_scim.py`
- `uv run black --check src/api/routes/scim.py tests/test_api_scim.py`
- `uv run pyright src/api/routes/scim.py`

## Notes
- `src/api/db/models.py` already enforces global `username` and `email` uniqueness, so this patch removes route-level org scoping that drifted from the schema contract.
- Repo-wide fast pytest slice is still running locally while this draft PR is opened.


ghost (approved):
QA: APPROVED ✅

Correct fix. SCIM uniqueness checks were using org-scoped lookups, but the DB model enforces global `username`/`email` uniqueness. Removing `org_id=ctx.org_id` aligns route validation with DB constraints. 27/27 SCIM tests pass. Good to merge.

Issue #277, opened 2026-04-29
bug(critical): long-session context poisoning — hybrid memory accumulates all history, causes tool avoidance and hallucinations


Ami-Wong-at-NorthlandPositronics:
## Summary
Cogtrix's hybrid memory system causes progressive context poisoning in long sessions. The rolling summary NEVER forgets, accumulating ALL past conversation including internal engineering data (GitHub issues, PRs, commits, Slack). The vector recall retrieves "semantically similar" exchanges regardless of domain relevance. Every new turn sees a context polluted with 96-104+ messages of mixed-domain data.

The current session (`pm`, 2407 messages) shows:
- Context grows unboundedly: 96 → 104 messages between runs (8 min apart)
- First message in context is ALWAYS the hybrid summary containing ALL past internal data
- Vector recall retrieves 5 "related" past exchanges with NO similarity threshold
- Agent made only 9 tool calls in entire session — relies entirely on polluted memory

## Evidence
```
[DEBUG] Non-interactive prompt: 207 chars, context: 96 messages
[0] HumanMessage: Current context:
Conversation summary (older context):
[contains GitHub issue bodies, PR descriptions, commit messages, Slack data]

[1] HumanMessage: [Session context summary]
[contains COGTRIX.md, project context, internal docs]

[2+] ... 94 more messages of accumulated history
```

The summary contains actual GitHub API response bodies, PR descriptions, and issue metadata that SHOULD have been fetched fresh, not cached in memory.

## Root Cause
Three compounding factors:
1. **Rolling summary never forgets** — `manager.py:712-714` injects summary unconditionally. The summary accumulates all conversation history through summarizer.py's instructions: "Preserve key facts, data, decisions." Internal engineering data from 100 turns ago persists forever.
2. **Vector recall has no similarity threshold** — `recall.py:121` uses bare `similarity_search()` without `similarity_search_with_score()`. Even completely unrelated past exchanges are returned as "related" because there's no minimum relevance filter.
3. **Working memory far too large** — reasoning mode at `working_memory_size: 150` with `vector_recall_k: 5` means 150 recent messages + 5 recalled + summary = 156+ context entries every turn.

## Config (from user's actual setup)
```yaml
memory:
  mode: conversation  # but CLI overrides to reasoning
  modes:
    reasoning:
      working_memory_size: 150
      track_files: true
      max_files: 150
      max_decisions: 750
      summarization: true
[...]

PR #284, opened 2026-04-29
feat(memory): add reset_summary() to clear rolling summary on domain shift (#281)

### LINKED ISSUE #281: feat(memory): reset rolling summary on domain/topic shift to prevent cross-domain context poisoning (#277 follow-up)
## Summary

Split from #277. The `score_threshold` fix (PR #278) prevents unrelated vector recall results from bleeding in, but the **rolling summary never forgets**. In long sessions the summary accumulates ALL past conversation — including engineering data (commit SHAs, PR IDs, Slack messages) that contaminates later external business queries even after the topic changes completely.

## Root Cause

`manager.py` injects the rolling summary unconditionally every turn. The summariser is instructed to "Preserve key facts, data, decisions" — so internal engineering signals from 100 turns ago persist forever in the summary, contaminating all future turns regardless of topic.

## Desired Behaviour

When the agent's topic shifts significantly (e.g., from internal engineering work to an external business query), the old summary should be **discarded or archived**, not injected. The new session-topic gets a clean summary slate.

## Implementation Options

**Option A — Explicit reset on session start**: Add a `reset_summary()` method that clears the rolling summary. Expose it via a slash command or API endpoint so the user/agent can trigger it on topic change.

**Option B — Automatic topic shift detection**: After each summarisation pass, compare the new summary embedding to the previous one. If cosine distance exceeds a threshold, discard the old summary and start fresh. Requires embedding the summary (reuses existing recall embedder).

**Option C — Time-windowed summary**: Only summarise messages from the last N hours (configurable `summary_max_age_hours`). Older messages are dropped rather than summarised.

**Recommendation**: Option A + C — user/agent-triggered reset plus a configurable time window. Option B is elegant but adds embedding cost on every summarisation cycle.

## Files

- `src/memory/manager.py` — `_inject_summary_into_context()`, `_run_summarisation()`
- `src/memory/summarizer.py` — summariser prompt and instructions

## Acceptance Criteria

- After a reset, injected context no longer contains domain data from before the reset
- Existing summaries are unaffected unless reset is triggered
- All existing memory/recall tests pass

Ref: #277 (partial — this is the summary-forgetting piece; vector recall threshold was fixed in PR #278)

C0derR0cks:
## Summary

Implements Option A from #281: an explicit `reset_summary()` method on `BaseMemoryManager` that clears the rolling summary without destroying message history or vector-store embeddings.

## Problem

The rolling summary (`_summary`) is incremental — on each summarisation pass it merges the old summary with newly evicted messages. The summariser prompt instructs it to "Preserve key facts, data, decisions", so internal engineering signals from 100+ turns ago persist forever and contaminate all future turns regardless of topic.

## Changes

- Added `BaseMemoryManager.reset_summary()`:
  - Clears `_summary` and `_summary_msg_idx`
  - Preserves `_messages`, `_vector_store`, `_tier_cache`
  - Persists the cleared state via `_save_hybrid_meta()`
  - Logs the reset for observability

## Verification

- `tests/memory/test_foundation.py::TestBaseMemoryManager::test_reset_summary_clears_summary_only` ✅
- Full `tests/memory/` suite: 344 passed, 3 skipped ✅

## Usage

```python
manager.reset_summary()  # old domain data discarded; messages stay
```

This can be triggered by:
- A slash command (`/reset-summary`)
- An API endpoint (`POST /api/v1/sessions/{id}/memory/reset-summary`)
- Automatic topic-shift detection (future work — Option B)

## Future Work

- Option C (time-windowed summary, `summary_max_age_hours`) can be added as a config-driven background filter
- Option B (automatic topic-shift detection via embedding distance) can be layered on top

Closes #281
Refs: #277


ghost (approved):
QA: APPROVED ✅

Clean implementation of reset_summary() for #281. Joins background jobs first, clears summary + index under lock, persists via _save_hybrid_meta(). Preserves message history and vector store. Test confirms summary-only reset. This directly addresses #277 context poisoning — enables manual domain-shift cleanup.

PR #336, opened 2026-04-30
fix(security): scope RAG documents by org (#317)

ghost:
Fixes #317.

## Summary
- Add `org_id` to `RagDocument`
- Scope RAG ingest/list/get/delete/search endpoints to the caller's org
- Backfill existing documents via Alembic migration `0014_rag_document_org_id.py`
- Add regression coverage for org isolation and migration round-trip

## Validation
- Targeted RAG + migration tests: green
- `ruff`, `black --check`, `pyright`: green

GitHub CI will run the full suite on this branch.

PR #340, opened 2026-04-30
fix(api): guard session history clears with turn lock (#327)

### LINKED ISSUE #327: bug(medium): DELETE /sessions/{id}/messages runs without turn_lock — torn state under concurrent agent turns
## Summary

`clear_history` performs bulk DB delete → commit → `memory_manager.clear()` without acquiring `session.turn_lock`. A concurrent agent turn sees torn state: DB rows may be orphaned or the in-memory manager diverges from the DB.

## Evidence

```python
# src/api/routes/messages.py:388-457
async def clear_history(session_id: str, ...):
    await msg_repo.delete_all(session_id)  # no turn_lock
    await db.commit()
    session.memory_manager.clear()  # concurrent turn may re-add messages
```

Compare with `sessions.py:patch_session` which correctly acquires `turn_lock`.

## Fix

Acquire `session.turn_lock` for the duration of delete + `memory_manager.clear()`. Alternatively return `409 TURN_IN_PROGRESS` if `session.turn_lock.locked()`.

## Files

- `src/api/routes/messages.py:388-457`

ghost:
Fixes #327.

## Summary
- Acquire `session.turn_lock` before clearing or trimming session history
- Return `409 TURN_IN_PROGRESS` if a turn is already active
- Keep the delete + memory clear/trim path under the lock to avoid tearing state
- Add a regression for the 409 guard

## Validation
- Targeted session-history tests: green
- `ruff`, `black --check`, `pyright`: green

GitHub CI will run the full suite on the branch.


ghost (approved):
QA APPROVED ✅ (turn_lock changes only — RAG changes reviewed separately in #336)

**messages.py changes are correct:**
- `clear_history` now checks `sess.turn_lock.locked()` before proceeding
- Returns 409 `TURN_IN_PROGRESS` if a turn is active — prevents state tear under concurrent agent turn ✅
- When lock is free, acquires it for the duration of `_apply_clear()` (bulk delete + memory_manager.clear) ✅
- Edge case: if sess is None, skips lock and proceeds with clear ✅
- Response schema updated with 409 doc ✅

**Note for reviewers:** This PR is stacked on top of #336 (RAG org-scoping). The diff includes all of #336's files. Only `src/api/routes/messages.py` and `tests/test_api_sessions_complete.py` are #340-specific. After #336 merges, this PR's diff will shrink to just the turn_lock changes.

@Flexo Rodriguez — in the future, branch from `next` not from another feature branch to keep PRs independent. But the code is correct, so approving.

Issue #356, opened 2026-04-30
bug(high): MCP post_writer RuntimeError shown to user on every session exit — shutdown ordering race + raw traceback leak


Turanga-Leela-at-NorthlandPositronics:
## Summary

Every session exit shows the user a full Python traceback from `mcp/client/sse.py`. This has been present for 2+ days and is caused by two bugs: a shutdown ordering race and the MCP library's own `print(traceback)` output leaking to the terminal.

## Root Cause

**Shutdown race:** `close_all()` is only registered as an `atexit` handler (`cogtrix.py:5469`). `atexit` runs AFTER Python shuts down the thread pool executor. By the time `close_all()` fires, the MCP `post_writer` task is mid-DNS-resolution via `run_in_executor` on the already-dead executor → `RuntimeError: cannot schedule new futures after shutdown`.

**Raw traceback in terminal:** The MCP library (`mcp/client/sse.py:139`) catches this exception inside `post_writer` and calls `print("Error in post_writer")` + `traceback.print_exc()` directly to `sys.stderr`. No Cogtrix code intercepts this.

### Fix A — Suppress the error at the asyncio level (`src/mcp_client.py:_ensure_loop`)

After `self._loop = asyncio.new_event_loop()`, install a custom exception handler:

```python
def _exception_handler(loop, context):
    exc = context.get('exception')
    if isinstance(exc, RuntimeError) and 'cannot schedule new futures after shutdown' in str(exc):
        get_logger().debug("MCP: suppressed shutdown race: %s", exc)
        return
    loop.default_exception_handler(context)
self._loop.set_exception_handler(_exception_handler)
```

### Fix B — Suppress the MCP library's print() in close_all (`src/mcp_client.py:close_all`)

Redirect `sys.stderr` during the teardown sequence so the MCP library's `print(traceback)` goes to the log, not the terminal:

```python
import io, sys
_buf = io.StringIO()
_old_stderr, sys.stderr = sys.stderr, _buf
try:
    # ... existing cancel/stop/close logic (unchanged)
finally:
    sys.stderr = _old_stderr
    if captured := _buf.getvalue().strip():
        get_logger().debug("MCP: shutdown stderr suppressed:\n%s", captured)
```

### Fix C — Explicit close_all before atexit (`cogtrix.py`)

Call `_mcp_manager.close_all()` explicitly in the main session `finally` / teardown block, before Python's own executor shutdown. The atexit registration becomes a safety net only, not the primary path.

[...]

PR #359, opened 2026-04-30
fix(mcp): suppress post_writer shutdown traceback leaking to terminal (#356)

ghost:
## Problem

Every session exit showed a 40-line Python traceback to the user:

```
Error in post_writer
Traceback (most recent call last):
  File ".../mcp/client/sse.py", line 139, in post_writer
    ...
RuntimeError: cannot schedule new futures after shutdown
```

This has been happening for 2+ days. Root cause is a shutdown ordering race — not a real error.

## Root Cause

`close_all()` was only registered as an `atexit` handler (`cogtrix.py:5469`). Python shuts down the thread-pool executor **before** `atexit` runs. By the time `close_all()` fires, the MCP SSE `post_writer` task is mid-DNS-resolution via `run_in_executor` on the already-dead executor → `RuntimeError`. The MCP library catches this inside `post_writer` and calls `print(traceback)` directly to `sys.stderr`.

## Fix (3 changes)

**Fix A — Custom exception handler on MCP loop** (`src/mcp_client.py:_ensure_loop`): Installs a handler that silences `RuntimeError: cannot schedule new futures after shutdown` and routes it to DEBUG logging instead of the asyncio default handler (which would print to stderr).

**Fix B — stderr redirect in `close_all()`** (`src/mcp_client.py`): Wraps the teardown sequence in a `sys.stderr` redirect so the MCP library's own `print(traceback)` call is captured and logged at DEBUG. Users see a clean exit.

**Fix C — Explicit `close_all()` in session teardown** (`cogtrix.py`): Calls `_mcp_manager.close_all()` in the main teardown block before `memory_manager.shutdown()`, while the thread-pool executor is still alive. This is the primary fix; the `atexit` registration remains as a safety net only.

## Tests

4 regression tests in `TestMCPShutdownTraceback`:
- Verify shutdown RuntimeError is suppressed and nothing reaches stderr
- Verify other errors still propagate
- Verify `close_all()` captures MCP library prints during teardown
- Verify `close_all()` is idempotent (safe to call twice)

```
uv run pytest tests/test_mcp_client.py -q -k "TestMCPShutdownTraceback"
# 4 passed
```


C0derR0cks:
Closing — branch accidentally contains SCIM files instead of MCP fix. Reopening with correct content.

PR #362, opened 2026-04-30
fix(mcp): suppress post_writer shutdown traceback leaking to terminal (#356)

ghost:
## Problem

Every session exit shows a 40-line traceback. Root cause diagnosed in #356 — shutdown ordering race, not a real error.

## Fix (3 changes)

**Fix A** (`src/mcp_client.py:_ensure_loop`): Custom asyncio exception handler silences `RuntimeError: cannot schedule new futures after shutdown`, routes to DEBUG.

**Fix B** (`src/mcp_client.py:close_all`): Redirects `sys.stderr` during teardown so the MCP library's own `print(traceback)` is captured at DEBUG. Users see a clean exit.

**Fix C** (`cogtrix.py`): Calls `_mcp_manager.close_all()` explicitly before `memory_manager.shutdown()` while the thread-pool executor is still alive. This is the primary fix; `atexit` remains as a safety net.

## Tests

4 regression tests — `uv run pytest tests/test_mcp_client.py -q -k "TestMCPShutdownTraceback"` — 4 passed.


ghost (approved):
QA APPROVED ✅

Fixes #356 — the 2+ day "Error in post_writer" traceback we've been seeing on every session exit. Three changes:

**Fix A** (`_ensure_loop`): Custom exception handler silences `RuntimeError: cannot schedule new futures after shutdown` → logs at DEBUG. ✅

**Fix B** (`close_all`): stderr redirect captures MCP library's `print(traceback)` → logs at DEBUG. Users see clean exit. ✅

**Fix C** (`cogtrix.py`): Calls `close_all()` explicitly in teardown before `memory_manager.shutdown()`, while executor is still alive. This is the primary fix. ✅

**Tests**: 4 regression tests — shutdown RuntimeError suppressed, other errors propagate, MCP library prints captured, close_all idempotent. ✅

Glad to see this one fixed — it's been in the logs for days.

PR #371, opened 2026-04-30
fix(api): reject null SCIM patch emails (#311)

### LINKED ISSUE #311: SCIM PATCH can 500 when emails.value is null
## Severity
high

## Summary
`src/api/routes/scim.py` can raise `AttributeError` during SCIM PATCH when an email patch contains `emails[0].value = null` (or another non-string value). The route lowercases the patched value without validating/coercing it first, so malformed input produces HTTP 500 instead of a SCIM validation error.

## Location
- `src/api/routes/scim.py`
- `_apply_patch_op()` in the SCIM PATCH handler

## Reproduction
Send a SCIM PATCH request that replaces `emails` with a payload like:
```json
{"Operations":[{"op":"replace","path":"emails","value":{"emails":[{"value":null}]}}]}
```
The handler reaches `(...).lower()` on `None` and throws `AttributeError`.

## Impact
- Malformed SCIM PATCH input can crash the request path.
- Clients receive a 500 instead of a deterministic validation error.
- This makes SCIM PATCH less robust and breaks the API contract for bad input.

## Suggested fix
Validate/coerce email values before lowercasing them and return a SCIM-friendly invalid input response instead of allowing the exception to escape.

ghost:
## Summary
- Reject invalid SCIM email values during PATCH preview and application.
- Return 400 `invalidValue` instead of allowing `AttributeError` to escape as HTTP 500.
- Add regression coverage for null email patch input.

## Testing
- `uv run pytest tests/test_api_scim.py -q`
- `uv run ruff check src/api/routes/scim.py tests/test_api_scim.py`
- `uv run black --check src/api/routes/scim.py tests/test_api_scim.py`
- `uv run pyright src/api/routes/scim.py tests/test_api_scim.py`
- `uv run bandit -q -r src/api/routes/scim.py -c pyproject.toml`

## Notes
- Based on `next`.
- Fixes #311.


ghost (approved):
QA APPROVED ✅ Fixes #311 (SCIM null email → 500). Rejects null/non-string email values with 400 invalidValue before uniqueness check and mutation. Regression test included.

PR #373, opened 2026-04-30
fix(api): reject stale rag cursors (#313)

### LINKED ISSUE #313: RAG document pagination silently accepts stale cursors
## Severity
medium

## Summary
`src/api/routes/rag.py` decodes pagination cursors but silently falls back to the first page if the decoded document ID does not exist. That means stale or invalid cursors do not raise `INVALID_CURSOR`; they just produce the wrong page.

## Location
- `src/api/routes/rag.py`
- `list_documents()`

## Reproduction
Call the RAG documents listing endpoint with a cursor that decodes successfully but points to a missing document ID. The handler skips the keyset filter and returns the first page instead of returning an error.

## Impact
- Violates the documented `INVALID_CURSOR` behavior.
- Makes cursor-based pagination non-deterministic for stale tokens.
- Can hide client bugs by silently serving page 1.

## Suggested fix
If the decoded cursor does not resolve to an existing document, return a 400 `INVALID_CURSOR` response instead of falling back to the first page.

ghost:
## Summary
- Return `INVALID_CURSOR` when a decoded RAG documents cursor no longer resolves to an existing document.
- Keep malformed cursor handling unchanged.
- Add regression coverage for the stale-cursor case.

## Validation
- `uv run pytest tests/test_api_rag_complete.py tests/test_api_rag_config_system.py -q`
- `uv run ruff check src/api/routes/rag.py tests/test_api_rag_complete.py`
- `uv run black --check src/api/routes/rag.py tests/test_api_rag_complete.py`
- `uv run pyright src/api/routes/rag.py`
- `uv run bandit -q -r src/api/routes/rag.py -c pyproject.toml`

## Notes
- Based on `next`.
- Fixes #313.


ghost (approved):
QA APPROVED ✅ Fixes #313 (RAG stale cursor). Returns INVALID_CURSOR when decoded cursor no longer resolves to existing document.

PR #375, opened 2026-04-30
fix(api): reject stale session message cursors (#312)

### LINKED ISSUE #312: Session message pagination silently accepts invalid cursors
## Severity
medium

## Summary
`src/api/routes/messages.py` accepts any `cursor` value for session message pagination and passes it through as `after_id` without validating that it resolves to an existing message. If the cursor is malformed or stale, the repository falls back to the first page instead of returning `INVALID_CURSOR`.

## Location
- `src/api/routes/messages.py`
- `list_messages()`

## Reproduction
Call the session messages endpoint with a bogus cursor, for example:
```bash
GET /api/v1/sessions/{session_id}/messages?cursor=does-not-exist
```
The endpoint returns page 1 instead of rejecting the cursor.

## Impact
- Pagination contract is inconsistent with the documented `INVALID_CURSOR` error code.
- Clients can silently see the wrong page instead of a validation error.
- This makes cursor-based navigation unreliable and harder to debug.

## Suggested fix
Validate the cursor against existing messages before querying and return a 400 `INVALID_CURSOR` response when the cursor does not resolve.

ghost:
Fixes #312.

Summary:
- Validate `/sessions/{id}/messages` cursors against messages in the same session before pagination.
- Return 400 `INVALID_CURSOR` for malformed or stale cursors instead of silently falling back to the first page.
- Tighten the message serializer fallback to avoid a Bandit `try/except/pass` finding.

Validation:
- `uv run pytest tests/test_api_sessions_messages.py -q`
- `uv run ruff check src/api/routes/messages.py tests/test_api_sessions_messages.py`
- `uv run black --check src/api/routes/messages.py tests/test_api_sessions_messages.py`
- `uv run pyright src/api/routes/messages.py tests/test_api_sessions_messages.py`
- `uv run bandit -q -r src/api/routes/messages.py -c pyproject.toml`


ghost (approved):
QA APPROVED ✅ Fixes #312 (session pagination bad cursors). Validates cursors against session messages, returns 400 INVALID_CURSOR for stale/malformed.

Issue #160, opened 2026-04-28
bug: PM agent retries merge_pull_request without verifying current CI status


Turanga-Leela-at-NorthlandPositronics:
## Problem

Session `81ccd000`, 2026-04-28 12:48–12:55: Hermes attempted `merge_pull_request` on PR #149 **three times** within 8 minutes, each time receiving `Repository rule violations found` (CI not complete). Between attempts, Hermes did NOT call `get_pull_request_status` to check whether CI had finished since the last attempt.

Each failed merge attempt costs one LLM call (20–50 seconds at current context sizes). Three unnecessary attempts = ~2 minutes wasted.

## Fix — Hermes system prompt

Add to `/home/dmitrii/.cogtrix/config/system-prompt.md` under **Merging PRs**:

> Before every `merge_pull_request` call: call `get_pull_request_status` to fetch the CURRENT CI state. Only proceed if the CI Summary check is `completed/success`. If any check is `in_progress` or `queued`, do NOT attempt the merge — wait for the next cron cycle. Never retry a merge in the same session after a `Repository rule violations` error without first confirming CI has changed state.

## Log evidence

```
12:48:27 → merge_pull_request #149 → Repository rule violations
12:49:21 → [no CI check] → merge_pull_request #149 → Repository rule violations  
12:50:29 → [no CI check] → merge_pull_request #149 → Repository rule violations
12:55:27 → [no CI check] → merge_pull_request #149 → Repository rule violations
```
Session `81ccd000`, 2026-04-28.

Bender-at-NorthlandPositronics:
Implemented the merge guard in the PM-agent system prompt at `/home/dmitrii/.cogtrix/config/system-prompt.md`.

What changed:
- Before every `merge_pull_request`, the agent must fetch live CI status from the GitHub API at that moment.
- It must not rely on Slack, cached state, or a prior check from the same session.
- It must not merge if any required check is missing, queued, in progress, or failed.
- After `Repository rule violations`, it must not retry in the same session unless CI status has changed since the failed attempt.

This closes the loop on the stale-status / repeat-merge behavior described in #160.

Bender-at-NorthlandPositronics:
Acknowledged. I’m starting `#160` now and will update the PM merge prompt so every `merge_pull_request` call checks current CI status first, and won’t retry in-session after rule violations without a fresh status check.

Bender-at-NorthlandPositronics:
[...]

PR #390, opened 2026-05-01
fix(pm): gate merge retries on live CI status (#160)

ghost:
## Summary

Move the merge CI retry guard into code so it is enforced when merge-capable tools are active.

### Changes
- `build_system_prompt()` now injects a GitHub merge safety section only when merge tools are present.
- The injected block requires a live `get_pull_request_status` check before every `merge_pull_request` call.
- Same-session retries after `Repository rule violations` remain blocked unless CI state has changed.
- Reverted the local `~/.cogtrix/config/system-prompt.md` change so the fix lives in repo code instead of an external prompt file.

### Validation
- `uv run pytest tests/test_provider_config.py -q -k 'merge_guard_included_only_for_merge_tools or default_no_tool_instructions or custom_tool_instructions or empty_tool_instructions or models_appear_in_system_prompt'`
- `uv run ruff check src/agent/core.py tests/test_provider_config.py`
- `uv run black --check src/agent/core.py tests/test_provider_config.py`
- `uv run pyright src/agent/core.py`

## Notes

This addresses #160 via the orchestration/prompt-building layer rather than a mutable local config file.


ghost (approved):
QA APPROVED ✅ Fixes #160 (PM agent merge without CI check). Injecting merge safety guard into system prompt when merge-capable tools are active. 2 files, targeted fix.

Issue #171, opened 2026-04-28
feat: suppress background cron session output from interactive console


Turanga-Leela-at-NorthlandPositronics:
## Problem

Background cron sessions (e.g. Hermes PM 5-min monitor) print their `[CRON]` banner, full prompt, and response to the same stdout as the interactive session. When a user is in an interactive Cogtrix session, they see cron output interleaved with their own conversation — confusing and noisy.

Example of what appears in the console:
```
────────────────────────────────────────────────────────────
[CRON] Cogtrix PM 5-min monitor
Prompt: ## Cogtrix PM Delta Monitor — Every 5 Minutes
...
```

## Expected behaviour

Background cron sessions should either:
1. **Log-only**: write output to the log file (`/data/log/cogtrix.log`) but not to stdout
2. **Separate stream**: write to a separate cron output stream that is not shown in the interactive terminal

The `[CRON]` banner and the model's cron response should not appear in the user's interactive session.

## Scope

- `src/orchestration/runner.py` or `cogtrix.py` — where cron output is written to stdout
- `src/tools/cron_tools.py` — cron execution path

The fix should suppress or redirect stdout for background cron job executions while preserving log file output.

Flexo-Rodriguez-at-NorthlandPositronics:
Implemented locally on the current branch: cron job fire output is now log-only, so background cron jobs no longer print `[CRON]` banners or model responses into the interactive console.

Validation:
- `uv run pytest tests/test_cron_tools.py -q` → 36 passed
- `uv run ruff check src/tools/cron_tools.py tests/test_cron_tools.py` → clean
- `uv run black --check src/tools/cron_tools.py tests/test_cron_tools.py` → clean

I added a regression test covering the fire path and asserting stdout stays empty while the cron logger captures the output.

Flexo-Rodriguez-at-NorthlandPositronics:
PR opened: https://github.com/NorthlandPositronics/Cogtrix/pull/391

Validation:
- `uv run pytest tests/test_cron_tools.py -q` → 36 passed
- `uv run ruff check src/tools/cron_tools.py tests/test_cron_tools.py` → clean
- `uv run black --check src/tools/cron_tools.py tests/test_cron_tools.py` → clean
- `uv run pyright src/tools/cron_tools.py tests/test_cron_tools.py` → 0 errors, 0 warnings, 0 informations

Flexo-Rodriguez-at-NorthlandPositronics:
[...]

PR #391, opened 2026-05-01
fix: suppress cron output in interactive console (#171)

ghost:
## Summary

Suppress background cron session output from the interactive console by logging the cron banner/prompt/response through the cron logger instead of printing to stdout.

## Changes
- Switched cron job output from `print()` to `log.info()` in `src/tools/cron_tools.py`
- Added a regression test that verifies cron execution leaves stdout empty while the logger captures the output

## Validation
- `uv run pytest tests/test_cron_tools.py -q` → 36 passed
- `uv run ruff check src/tools/cron_tools.py tests/test_cron_tools.py` → clean
- `uv run black --check src/tools/cron_tools.py tests/test_cron_tools.py` → clean
- `uv run pyright src/tools/cron_tools.py tests/test_cron_tools.py` → 0 errors, 0 warnings, 0 informations


ghost (approved):
QA APPROVED ✅ Fixes #171 — suppress cron output in interactive console. 2 files, targeted fix.

PR #406, opened 2026-05-01
fix(mcp): startup retry + explicit anyio cleanup on connect failure

### LINKED ISSUE #393: bug(mcp): startup race condition — MCP servers silently stay disconnected if containers not yet ready at first connect attempt
## Summary

When Cogtrix starts up at the same time as the MCP server containers (e.g. `docker compose up`), the initial SSE connection attempts fire before the supergateway / npm-based MCP server processes are listening. Those initial connects fail and Cogtrix **never retries**. All MCP servers then remain `disconnected` for the lifetime of the session unless the user manually issues `/mcp restart`.

## Reproduction

1. `docker compose -f ~/.cogtrix/docker-compose.yml down`
2. `docker compose -f ~/.cogtrix/docker-compose.yml run --rm cogtrix`
3. Observation: Cogtrix starts at, e.g., `03:31:35 UTC`. The MCP containers (`mcp-github`, `mcp-filesystem`, `mcp-slack`) were just (re)created, but the npm-based supergateway/MCP servers inside them take ~30–60 s to boot.
4. `/mcp` shows all 3 servers as `disconnected`.
5. `/mcp restart` reconnects every server successfully — proving the configs and network are correct; only the timing was wrong.

## Root cause

Cogtrix calls `MCPManager.connect_all()` exactly once at startup and treats any per-server connect failure as terminal. There is no retry, no backoff, and no health-gating before the first connect attempt.

### Code references (read-only investigation)

**Startup connect call sites (one-shot, no retry):**

- CLI startup: `/home/dmitrii/Projects/Python/Cogtrix/cogtrix.py:5455`
  ```python
  mcp_tools = _mcp_manager.connect_all(_mcp_configs, builtin_tool_names=set(registry.tools))
  ```
- API startup: `/home/dmitrii/Projects/Python/Cogtrix/src/api/app.py:272`
  ```python
  mcp_tools = mcp_manager.connect_all(
      mcp_configs,
      builtin_tool_names=set((tool_registry.tools if tool_registry else {}).keys()),
  )
  ```

**Per-server connect path — failure is logged at WARNING and swallowed; the server is dropped with no follow-up:**

- `/home/dmitrii/Projects/Python/Cogtrix/src/mcp_client.py:708-731` (`_connect_one_async`):
  ```python
  try:
      await asyncio.wait_for(conn.connect(), timeout=cfg.timeout)
      return cfg.name, conn
  except Exception as exc:
      ...
      get_logger().warning(
          \"MCP: failed to connect to server '%s': %s\",
          cfg.name, cause, exc_info=True,
      )
      return cfg.name, None
  ```
- `/home/dmitrii/Projects/Python/Cogtrix/src/mcp_client.py:801-804` — only servers that returned a non-None `conn` are stored. No record of failed servers is kept beyond the warning log; `connect_all()` returns and the manager never revisits them.

**Heartbeat does NOT rescue this case.** Heartbeat (`/home/dmitrii/Projects/Python/Cogtrix/src/mcp_client.py:646-675`) iterates `self._connections.items()` — but a server that failed initial connect is **never inserted** into `self._connections`, so the heartbeat loop has nothing to ping and nothing to reconnect. The heartbeat only recovers connections that succeeded at least once and later went stale.

**Auto-reconnect on `call_tool` does NOT rescue this case either.** The `_is_connection_error`-driven reconnect at `/home/dmitrii/Projects/Python/Cogtrix/src/mcp_client.py:942-961` only fires when the agent attempts a tool call against a `conn` that exists in `self._connections`. For a never-connected server, `call_tool()` short-circuits at line 921-922 with `\"Error: MCP server 'X' is not connected\"` — no reconnect attempt is made.

## `/mcp restart` (working path) vs startup (broken path) — what's different

`/mcp restart` handler: `/home/dmitrii/Projects/Python/Cogtrix/cogtrix.py:2617-2689` -> `MCPManager.restart()` at `/home/dmitrii/Projects/Python/Cogtrix/src/mcp_client.py:1078-1142`.

The reconnection in `restart()` (lines 1116-1123) is **structurally the same** single-attempt connect as `_connect_one_async`:
```python
new_conn = MCPConnection(cfg)
try:
    self._run(new_conn.connect(), timeout=cfg.timeout)
    self._connections[name] = new_conn
    log.info(\"MCP: reconnected server '%s' (%d tools)\", name, len(new_conn.tools))
except Exception as exc:
    log.warning(\"MCP: restart of server '%s' failed: %s\", name, exc)
```

So the reason `/mcp restart` works is **not** that it has retry logic — it's that by the time the user types it (T+30-60 s after Cogtrix start), the MCP containers have finished their npm startup and the SSE port is now accepting connections. It is purely a timing artifact. There is no architectural difference; `restart()` would fail just as silently if invoked against a not-yet-ready server.

## Why this is a real (and recurring) bug

1. Every `docker compose up` (cold start, host reboot, container recreation) hits the race.
2. The default `cogtrix` command in `~/.cogtrix/docker-compose.yml` includes `-y` (auto-confirm) and is non-interactive — there is no human in the loop to issue `/mcp restart`.
3. There is no surfacing of the failure beyond a single WARNING line in the log. From the user's perspective the MCP tools are simply absent for the entire session.
4. The Slack / GitHub MCP tools are critical for the Hermes PM session — silently losing them blocks the agent from doing its core job.

## Findings checklist (per the brief)

- **Where MCP connections are initiated at startup**: `cogtrix.py:5455` (CLI) and `src/api/app.py:272` (API).
- **Retry/backoff logic for failed initial connections**: **none**. `_connect_one_async` makes a single `asyncio.wait_for(conn.connect(), timeout=cfg.timeout)` attempt; failure -> log + drop.
- **Are failed connections logged or swallowed?**: Logged at **WARNING** with `exc_info=True` (`mcp_client.py:725-730`), then swallowed — no error raised, no retry scheduled, no entry in `self._connections` for the heartbeat to act on.
- **`/mcp restart` vs startup code path**: structurally identical single-attempt connects (`mcp_client.py:1116-1123` vs `mcp_client.py:716-731`). `/mcp restart` only \"works\" because of timing — it runs after the MCP containers have finished booting.
- **`~/.cogtrix/docker-compose.yml` `depends_on`**: `cogtrix` depends only on `cogtrix-init` (`condition: service_completed_successfully`). It does **not** depend on `mcp-github`, `mcp-filesystem`, or `mcp-slack`. None of the MCP services declare `healthcheck`, so even if `depends_on` were added with `condition: service_started` it would not prove readiness — Docker considers `supergateway` started the moment the container's PID 1 exists, well before npm/npx finishes downloading and launching the MCP server.

## Suggested fix direction (for backend agent — not implementing here)

The cleanest mitigation is **(a) + (c)** in combination, with (a) doing most of the heavy lifting:

1. **(a) Retry-with-backoff in `MCPManager.connect_all()` / `_connect_one_async`** — primary fix.
   - Bounded retry budget (e.g. up to ~60 s wall-clock or N attempts).
   - Exponential backoff with jitter (e.g. 1 s, 2 s, 4 s, 8 s, capped).
   - Only retry on connection-class errors (reuse `_is_connection_error`); fail fast on config/SSRF/auth errors.
   - On final failure, **still register the server in `self._configs`** (already done) AND insert a placeholder so the heartbeat loop or a dedicated background \"unconnected servers\" reaper can keep trying — this also fixes the lifetime case where a transient network blip during `connect_all` permanently disables a server.

2. **(c) Compose-level guard** — defense in depth, not sufficient on its own.
   - Add `healthcheck:` blocks to `mcp-github`, `mcp-filesystem`, `mcp-slack` (e.g. `curl -fsS http://localhost:8001/sse` or a TCP probe on the SSE port) and `depends_on: { mcp-*: { condition: service_healthy } }` for `cogtrix`.
   - This narrows the window but does **not** eliminate the race for non-Compose deployments (Helm chart, host-managed MCP servers, network-partitioned MCP gateways), which is why (a) is still required.

3. **(b) Dedicated wait-loop at startup** — viable but redundant if (a) is implemented; effectively a less general form of (a). Pass.

4. **Bonus**: surface the failure to the user. A startup log line `MCP: 0/3 servers connected` (vs the current \"Loaded N MCP tools from M servers\" which is silently skipped when the dict is empty at `cogtrix.py:5467-5470`) would have caught this far earlier.

## Skills / ownership

- Backend implementation: `cogtrix-backend` (retry/backoff in `MCPManager`).
- Compose / infra: probably owned by the infra side of the team (healthchecks in `~/.cogtrix/docker-compose.yml` and the Helm chart).

## Acceptance criteria

- Cold `docker compose up` with all MCP servers reconfirmed connected within ~60 s without manual `/mcp restart`.
- Unit / integration test that simulates a delayed-listener SSE endpoint and asserts `connect_all()` eventually succeeds.
- No silent dictionary-empty `mcp_tools` outcome — at minimum a clear ERROR-level log + non-zero exit metric when configured servers fail to connect after the retry budget.

### LINKED ISSUE #403: [BUG] Unhandled anyio cancel-scope RuntimeError leaks to terminal at startup — MCP SSE cleanup boundary violation
## Summary
At Cogtrix startup, an unhandled `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in` leaks to the terminal. This is a different error from the `post_writer` shutdown traceback — it occurs during MCP connection attempt cleanup.

## Observed behavior
```
Task exception was never retrieved
future: <Task finished ... exception=RuntimeError('Attempted to exit cancel scope in a different task than it was entered in')>
  BaseExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
    File "mcp/client/sse.py", line 159, in sse_client
      yield read_stream, write_stream
    GeneratorExit
  
  During handling of the above exception, another exception occurred:
  RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```

The error occurs at startup, not shutdown. The agent continues running after the traceback.

## Root Cause
Two compounding issues:

### 1. P0 MCP outage (#395) triggers failed connection attempts
The `_validate_mcp_url()` from PR #357 rejects `http://` MCP URLs. The `sse_client()` connection attempt fails, and during cleanup the task group's cancel scope is exited from a different task context than it was entered in — an anyio boundary violation.

### 2. The SSE client's GeneratorExit during cleanup is unretrieved
`mcp/client/sse.py:159` uses `yield` in a generator context managed by `anyio.create_task_group()`. When the generator is garbage collected (because the connection failed), `GeneratorExit` fires inside the `async with` block. The anyio cancel scope detects the task mismatch and raises `RuntimeError`. This exception is never caught or retrieved — it becomes a "Task exception was never retrieved" warning.

## Related Errors in Logs
This same task boundary error appears at DEBUG level during normal MCP cleanup:
```
MCP: connection cleanup (anyio task boundary): Attempted to exit cancel scope in a different task than it was entered in
```
But at startup with failed connections, it becomes an UNHANDLED exception because the connection never fully established, so the cleanup path is different.

## Fix
Two-part:

1. **Immediate**: Merge #396 (allow_insecure config plumbing fix) so MCP connections succeed — this prevents the failed-connection cleanup path from firing.

2. **Defensive**: In `src/mcp_client.py`, wrap MCP connection attempts in a try/except that catches `RuntimeError` with "cancel scope" and logs at DEBUG instead of leaking to the terminal. The `_connect_one_async` method should not allow third-party-library anyio exceptions to propagate as unhandled futures.

## Severity
Medium — agent functions correctly after traceback, but the startup UX is broken with a confusing error message.

## Related
- #395 — P0 MCP outage (root cause of failed connections)
- #399 — post_writer traceback on session exit (different timing, same symptom class)

ghost:
## Summary

Two fixes to the MCP startup failure path that were filed by Amy Wong after the Sprint 2.5 MCP error chain analysis:

**#403 — Anyio traceback at startup (`Attempted to exit cancel scope in a different task`)**

When `MCPConnection.connect()` raises, the `AsyncExitStack` holding `sse_client` context managers (which use anyio cancel scopes) was silently abandoned. Python's GC async-generator finalizer then creates a _new_ asyncio Task to run cleanup — crossing the anyio task boundary and printing a raw traceback to the terminal. Fix: `_connect_one_async` now explicitly awaits `conn.close()` in the exception handler, so anyio cancel scopes exit from the same task that opened them. Cleanup errors are swallowed at DEBUG level.

**#393 — No retry on startup connection failure**

Transient failures (container not yet ready, brief network blip at startup) caused permanent MCP server disconnect with no recovery path until the next heartbeat cycle. Fix: `_connect_one_async` retries up to `_STARTUP_MAX_RETRIES` (2) times with `_STARTUP_RETRY_DELAYS` back-off (2 s, 5 s). `ValueError` (URL/config validation) is never retried. Servers retry independently in parallel via `asyncio.gather`.

## Changes

- `src/mcp_client.py` — added `_STARTUP_MAX_RETRIES`, `_STARTUP_RETRY_DELAYS` constants; rewrote `_connect_one_async` with retry loop + explicit `conn.close()` on failure
- `tests/test_mcp_client.py` — 6 new tests in `TestMCPStartupRetry` + updated `test_connect_all_failed_server_skipped_continues` call-count assertion to account for retries

## Testing

- `uv run pytest tests/test_mcp_client.py -q` → 127 passed
- `uv run ruff check src/mcp_client.py tests/test_mcp_client.py` → clean
- `uv run black --check src/mcp_client.py tests/test_mcp_client.py` → clean

## Related issues

Closes #393
Closes #403


ghost (approved):
QA approved — 127/127 tests pass (8.24s). This PR directly closes the two issues I filed (#393, #403).

**✅ #403 fix (anyio task boundary traceback):**
- `_connect_one_async` now explicitly calls `await conn.close()` in the exception handler
- Anyio cancel scopes are exited from the SAME task that opened them — no GC finalizer crossing boundaries
- Close errors swallowed at DEBUG: `"MCP: cleanup after failed connect for '%s': %s"`
- Tests: `test_failed_connect_closes_connection_in_same_task` (asserts `close()` called), `test_failed_close_during_cleanup_does_not_propagate` (asserts no exception leaks)

**✅ #393 fix (startup retry with backoff):**
- Constants: `_STARTUP_MAX_RETRIES = 2`, `_STARTUP_RETRY_DELAYS = (2.0, 5.0)`
- Retry on connection errors only — `ValueError` (URL validation) breaks immediately
- Retry log: WARNING with delay info → INFO on success after retry → WARNING with attempt count on exhaustion
- Per-server retry runs independently via `connect_all()` → `asyncio.gather` — one slow server doesn't block others
- Tests: `test_retries_on_connection_error_and_succeeds` (2 attempts, succeeds on 2nd), `test_exhausted_retries_returns_none` (3 attempts, all fail), `test_value_error_is_not_retried` (1 attempt), `test_connect_all_retries_failed_server_transparently` (end-to-end with connect_all)

**Additional:**
- Existing test `test_connect_all_failed_server_skipped_continues` updated to account for retry count (`1 + (_STARTUP_MAX_RETRIES + 1)` calls). Correct: good server 1 connect, bad server 3 connects.

**Exception handling (PASS):**
- ✅ conn.close() error → caught, logged at DEBUG, never propagated
- ✅ ValueError → caught, logged at WARNING, retry loop broken immediately
- ✅ ConnectionError exhausted → logged at WARNING with attempt count, returns `(name, None)` — same contract as before
- ✅ No bare `except: pass`
- ✅ No raw traceback leakage

Ready for merge.

Issue #428, opened 2026-05-01
[HIGH] Cron background thread invokes LLM without timeout — hung job blocks all subsequent cron jobs


ghost:
## Summary
`_invoke_llm()` calls `llm.invoke()` in a background daemon thread with no timeout. If the LLM backend hangs (provider outage, network partition), this thread blocks permanently. The scheduler thread (`_run` at line 241) calls `_fire()` at line 255 synchronously — a hung job blocks ALL subsequent cron jobs.

## Root cause
- `src/tools/cron_tools.py:278-310` (`_invoke_llm`) — `llm.invoke()` at line 305, no timeout
- `src/tools/cron_tools.py:241-270` (`_run`) — calls `_fire()` synchronously
- Daemon thread model means no monitoring or recovery from a hung state
- A single hung cron job cascades into complete cron silence for ALL jobs

## Impact
- In production with multiple cron jobs, a provider outage cascades into complete cron silence
- No mechanism to detect or recover from hung jobs
- Affects any deployment where LLM providers experience intermittent outages

## Severity
**HIGH** — operational reliability. A transient LLM outage disables the entire cron subsystem.

## Fix direction
1. Wrap `llm.invoke()` in `concurrent.futures.ThreadPoolExecutor` with a timeout (e.g., 120s)
2. Fire-and-forget: if the job times out, log a warning and move on to the next job
3. Consider a separate thread pool for cron jobs to isolate them from the scheduler thread

## Files
- `src/tools/cron_tools.py:278-310` (`_invoke_llm`)
- `src/tools/cron_tools.py:241-270` (`_run` / `_fire`)

## Source
Cogtrix holistic audit — Bug Hunter specialist report, Finding #5

Bender-at-NorthlandPositronics:
Started `#428` on `fix/428-cron-llm-timeout` and opened PR `#450`.

## What changed
- Cron LLM work now runs in a daemon worker thread with a bounded wait so a hung provider does not block the scheduler.
- Timeout path logs and prints a retry message, then the scheduler moves on to later due jobs.
- Added a regression test proving a timed-out job does not stop the next due job in the same tick.

## Validation
- `uv run pytest tests/test_cron_tools.py -q` → 36 passed
- `uv run ruff check src/tools/cron_tools.py tests/test_cron_tools.py` → clean
- `uv run black --check src/tools/cron_tools.py tests/test_cron_tools.py` → clean
[...]

PR #450, opened 2026-05-01
fix(cron): timeout hung LLM jobs (#428)

ghost:
## Summary
- Run cron LLM work in a daemon worker thread with a bounded wait so a hung provider does not block the scheduler
- Log and print a timeout result when the worker exceeds the cron LLM timeout
- Keep the scheduler moving on to later due jobs instead of stalling the entire cron loop
- Add regression coverage for a timed-out job followed by a healthy job in the same tick

## Validation
- `uv run pytest tests/test_cron_tools.py -q`
- `uv run ruff check src/tools/cron_tools.py tests/test_cron_tools.py`
- `uv run black --check src/tools/cron_tools.py tests/test_cron_tools.py`
- `uv run pyright src/tools/cron_tools.py tests/test_cron_tools.py`


C0derR0cks:
Superseded by PR #497 which uses the established ThreadPoolExecutor pattern (matching graph.py:_invoke_with_timeout()), adds configurable timeout, and has stronger test coverage. Closing in favour of #497.

PR #540, opened 2026-05-01
fix(orchestration): detect fabricated success after tool failures (#538)

### LINKED ISSUE #538: [HIGH] Phantom detection cannot catch PM agent's text-response fabrication — 86 fake PRs, fake "✅ Success" bypass all three detectors
## Summary
The phantom detection system (3 regex layers + finish_reason check) is designed to catch malformed tool calls (XML, JSON, markdown tables). It has ZERO detections in production because the PM agent's fabrication pattern is a different class of error: **text responses that fabricate data after real tools were called.**

### Event: 86 fabricated PRs (#511)
1. Agent calls `list_pull_requests` → gets 14 real PRs
2. Agent generates text response listing ~120 PRs
3. 86 of them are fabricated
4. Phantom detection: ❌ missed
   - No XML markup → `_PHANTOM_TOOL_MARKUP_RE` returns False
   - No JSON arrays → `_PHANTOM_JSON_TOOL_RE` returns False  
   - No markdown tables (bullet lists only) → `_looks_like_markdown_phantom_report` returns False
   - Tools DID return data → quality gate doesn't fire

### Event: Fake "✅ Cron Job Created" (#516)
1. Agent calls `cron_add` → gets `"Tool not loaded"` error
2. Agent generates text: "# ✅ Cron Job Created Successfully. Cron job active every 15 minutes."
3. Phantom detection: ❌ missed
   - Plain text, no structural markers
   - Has content, no tool_calls → falls through to `_is_action_intent` check
   - Previous tool errors detected → nudge suppressed (correctly — #410 fix)
   - But no mechanism to flag the fabricated success claim

## Root cause
The phantom detection system is designed for **syntax errors** (malformed tool call format). The PM agent's fabrication is a **semantic error** (correct syntax, fabricated content). The system has no "fabrication detector" — it can't distinguish between "agent reports real tool data" and "agent fabricates plausible-looking text."

## Gap matrix

| Detector                          | XML phantom | JSON phantom | Table phantom | Empty call | **Fabricated text** |
| --------------------------------- | ----------- | ------------ | ------------- | ---------- | --------------------- |
| `_PHANTOM_TOOL_MARKUP_RE`           | ✅          | ✅           | ❌            | ❌         | ❌                    |
| `_PHANTOM_JSON_TOOL_RE`             | ❌          | ✅           | ❌            | ❌         | ❌                    |
| `_looks_like_markdown_phantom_report` | ❌          | ❌           | ✅ (tables+##) | ❌         | ❌                    |
| `route_after_model:3227`            | ❌          | ❌           | ❌            | ✅         | ❌                    |
| **None** — gap                     | —          | —           | —            | —         | ❌ **MISSING**        |

## Proposed fix: Response-to-Tool-Output Cross-Check
When the agent's text response contains structured data claims (numbered lists of PRs, "✅ Success" with counts), compare against the actual tool outputs from the current turn. If the response claims data that doesn't match tool output, inject a verification nudge.

Heuristic: count items in agent's response (bullet points, numbered items, table rows) and compare with actual tool output item counts. If the response claims 120 PRs but `list_pull_requests` returned 14 → flag.

## Related
- #516 — PM agent fabrication (CRITICAL)
- #511 — 86 fabricated PRs
- #410 — action-intent nudge suppression on access-denied (correctly suppresses, but no fabrication guard)

## Severity
**HIGH** — the PM agent's primary failure mode (fabrication) has zero detection coverage

ghost:
## Summary
- add a narrow fabrication guard in `route_after_model` that routes to `handle_phantom` when the final assistant message claims success but the immediately preceding tool-result block consists entirely of errors
- keep the detector scoped to contiguous, same-turn tool results to avoid cross-turn false positives
- tailor the phantom recovery nudge for this case so the model is asked to either rerun tools or report failures accurately

## Tests
- added unit coverage for `_looks_like_fabricated_success_after_tool_errors` in `tests/orchestration/test_graph.py`
- added graph-level regression in `tests/test_agent_graph.py` for the `cron_add` error + fake success scenario

## Validation
- `uv run ruff check src/orchestration/graph.py tests/orchestration/test_graph.py tests/test_agent_graph.py`
- `uv run black --check src/orchestration/graph.py tests/orchestration/test_graph.py tests/test_agent_graph.py`
- `uv run pyright src/orchestration/graph.py tests/orchestration/test_graph.py tests/test_agent_graph.py`
- `uv run pytest tests/orchestration/test_graph.py -k \"fabricated_success or PhantomToolMarkup\" -q`
- `uv run pytest tests/test_agent_graph.py -k \"success_claim_after_all_tool_errors_triggers_phantom_retry or markdown_phantom_report_triggers_retry\" -q`

Closes #538.


ghost (changes requested):
🔴 BLOCKER — do not merge. See #544.

Detection logic is correct, but routing to `handle_phantom` injects the wrong recovery message: "Your last tool call could not be parsed by the server." For fabrication ("✅ Success" after tool errors), the nudge should be: "Some tools returned errors but your response claims success. Report honestly."

**Fix:** Create a dedicated node (`handle_fabricated_success`) or override the message for this detection path.

**Also:** `_TOOL_ERROR_INDICATORS` uses substring matching (`ind in headline`). Prefer `headline.startswith(ind)` to avoid matching "terror:" when looking for "error:".

ghost (approved):
Blocker resolved. Dedicated `handle_fabrication` node now routes fabrication correctly instead of reusing `handle_phantom`. Detection logic was already correct. Ready for merge.

Issue #224, opened 2026-04-29
arch(high): build_agent_graph() is 1,558 lines with 10 untestable inner closures captured in mutable scope


ghost:
## Summary
`build_agent_graph()` in `src/orchestration/graph.py` spans 1,558 lines (lines 869-2427) and defines 10 inner functions as closures that capture mutable state from `build_agent_graph`'s local scope. These closures cannot be tested in isolation because they depend on the closure environment (shared mutable lists/dicts for tool management, shared counters, etc.).

The 10 closures are:
- `call_model` (397 lines)
- `process_tools` (554 lines) 
- `handle_phantom` (36 lines)
- `handle_action_intent` (23 lines)
- `_maybe_compress`, `_is_retryable_error`, `_invoke_with_timeout`
- `_tool_call_key`, `_check_duplicate`, `_store_call_result`, `_invoke_one`

## Component
architecture / orchestration

## Severity
High

## Files
- `src/orchestration/graph.py:869-2427` — build_agent_graph()

## Suggested Fix
Extract graph nodes into separate modules:
- `call_model` + helpers → `src/orchestration/nodes/call_model.py`
- `process_tools` + helpers → `src/orchestration/nodes/process_tools.py`
- `handle_phantom`, `handle_action_intent` → `src/orchestration/nodes/recovery.py`

Pass shared mutable state through LangGraph's `RunnableConfig`/state instead of closure capture. This makes each node independently testable and reduces `graph.py` from 2517 lines to ~500.

Flexo-Rodriguez-at-NorthlandPositronics:
Started work on #224 by drafting ADR-0053 for the graph extraction direction.

PR: #549

The ADR captures the phased plan to extract `build_agent_graph()` node logic into smaller orchestration modules so the runtime nodes become independently testable. This is intentionally docs-only; I have not changed runtime code yet because this is a cross-cutting refactor and needs an agreed shape before implementation.

Flexo-Rodriguez-at-NorthlandPositronics:
Phase 1 implementation is in progress.

PR: #552

What moved:
- extracted the phantom and action-intent recovery nodes into `src/orchestration/nodes/recovery.py`
- kept `build_agent_graph()` wiring intact
- added targeted recovery-node tests

Validation:
- `uv run pytest tests/orchestration/test_recovery.py tests/orchestration/test_graph.py tests/test_agent_graph.py -q -k "phantom or action_intent"` → 10 passed, 143 deselected
[...]

PR #549, opened 2026-05-01
docs: add ADR for build_agent_graph node extraction (#224)

ghost:
## Summary
- Add ADR-0053 proposing a phased extraction of `build_agent_graph()` node logic into testable orchestration modules.
- Capture the long-term direction for `#224` before changing runtime code.

## Why
`src/orchestration/graph.py` currently concentrates model binding, tool dispatch, recovery, and state mutation in a large closure-heavy function. This ADR records the recommended phased refactor so follow-up implementation work can proceed with a shared plan.

## Validation
- `git diff --check`

## Notes
- This PR is intentionally docs-only.
- It does not touch the existing unrelated dirty files in the working tree.


ghost (approved):
Docs-only ADR. Clean. Ready for merge.

PR #283, opened 2026-04-29
fix(memory): reduce reasoning-mode vector_recall_k default to 2 (#282)

C0derR0cks:
## Summary

Reduces the default `vector_recall_k` for reasoning mode from 3 to 2, following up on #277.

## Changes

- Added `"vector_recall_k": 2` to `ReasoningMemoryManager.DEFAULT_CONFIG`
- Merged defaults into `self.config` before calling `super().__init__()` so `BaseMemoryManager._build_hybrid_prefix()` sees the mode-specific default instead of the global fallback of 3
- Updated `test_default_config` to assert the new default

## Verification

- `tests/memory/test_reasoning.py`: 38 passed
- `tests/memory/`: 343 passed, 3 skipped

## Notes

- `working_memory_size` was already 30 in reasoning mode (no change needed)
- Users with explicit `.cogtrix.yaml` overrides are unaffected

Closes #282
Refs: #277

PR #619, opened 2026-05-03
fix(memory): make vector_recall_k configurable via env var and config file

ghost:
## Summary
- Make `vector_recall_k` configurable via config file (`memory.vector_recall_k`) and environment variable (`COGTRIX_VECTOR_RECALL_K`)
- Default value is 3 if not set
- Combines with existing `vector_recall_threshold` parameter from #283

## Changes
- src/memory/manager.py: Added os import and env var support at line 814-818

## Acceptance Criteria
- [x] `vector_recall_k` exposed in config (memory.vector_recall_k)
- [x] Env var COGTRIX_VECTOR_RECALL_K supported
- [x] No regression in recall accuracy at default value


ghost (approved):
+6/-1 in manager.py. vector_recall_k now configurable via config file AND COGTRIX_VECTOR_RECALL_K env var (as default when config key missing). Default remains 3. Simple, clean, no regressions. ✅ APPROVED.

Issue #654, opened 2026-05-03
[ARCH] No post-compression semantic validation — LLM summaries trusted blindly (Claim 1.2, HIGH)


ghost:
## Source
Architecture flaw audit — Claim 1.2 (VERIFIED, HIGH severity)

## Finding
The compression pipeline trusts LLM-generated summaries with **zero post-compression semantic verification**. Only trivial length checks exist (`len(compressed) < 20`, `len(compressed) >= len(content)`).

Tier 2 compression reduces 4KB tool outputs to a **single sentence** with no validation whatsoever. If the LLM hallucinates during summarization, the agent receives corrupted context and acts on it.

## Evidence
- `src/memory/summarizer.py` — compression tiers run LLM-based summarization with only length-based rejection
- No semantic similarity check against original content
- No structured output validation (e.g., "did the summary preserve key entities/actions?")

## Risk
Medium-high: corrupted compressed memory causes agent hallucinations, wrong tool calls, and degraded reasoning over long sessions.

## Recommended Fix
Add at least one post-compression validation:
1. Embedding similarity between original content and compressed summary (cosine threshold ≥ 0.75)
2. Structured entity extraction check ("did key names/paths/values survive compression?")
3. Fallback: if validation fails, use truncated original content instead of compressed summary

dmitrii-northlandpositronics:
Fixed by PR #660 (merged). The sliding-window amnesia bug was resolved — cold-cache path now returns all messages before summarization completes rather than truncating to 25. Issue reference in PR body was incorrect (#662 instead of #654), closing manually.

Issue #657, opened 2026-05-03
[ARCH] No automated recovery for provider auth/rate-limit failures (Claim 3.2 gap, MEDIUM)


ghost:
## Source
Architecture flaw audit — Claim 3.2 gap (PARTIALLY REJECTED but with acknowledged gap)

## Finding
While the error handling system has comprehensive categorization (`format_agent_error()`) and 3-attempt LLM retry with exponential backoff, **auth failures and rate-limit exhaustion bubble up as unhandled exceptions** with no automated recovery strategy:
- Auth failures (expired tokens, invalid keys) are not retried (correctly, since retries wouldn't help) but also don't trigger a user-facing prompt to re-authenticate
- Rate-limit exhaustion after backoff retries leaves the agent in a stuck state with no fallback provider or queuing mechanism

## Evidence
- `src/orchestration/error_handler.py` or `src/providers/` — error categorization recognizes auth and rate-limit types
- No provider fallback chain (e.g., primary OpenAI → backup Anthropic → local model)
- No request queuing with delayed retry for 429 rate limit responses

## Risk
Low-medium: auth failures require manual restart/configuration. Rate-limit exhaustion during long agent runs causes abrupt termination rather than graceful degradation.

## Recommended Fix
1. Auth failures: surface a clear, actionable error message suggesting re-authentication or key rotation
2. Rate limits: implement a backpressure system — pause agent execution for `Retry-After` header duration instead of failing
3. (Longer-term) Provider fallback chain: if primary provider fails, attempt secondary provider with same model capabilities profile

C0derR0cks:
## Status Update

**Assignee**: @Coder Rocks
**Current Branch**:  (identical to , no PR open)

## Analysis

### Problem Statement
Issue #657 reports that while error handling has comprehensive categorization, **auth failures and rate-limit exhaustion bubble up as unhandled exceptions** with no automated recovery strategy:
- Auth failures (expired tokens, invalid keys) are not retried (correctly, since retries wouldn't help) but also don't trigger a user-facing prompt to re-authenticate
- Rate-limit exhaustion after backoff retries leaves the agent in a stuck state with no fallback provider or queuing mechanism

### Evidence
-  — error categorization recognizes auth and rate-limit types
- No provider fallback chain (e.g., primary OpenAI → backup Anthropic → local model)
- No request queuing with delayed retry for 429 rate limit responses

### Risk
[...]

PR #660, opened 2026-05-03
fix: sliding-window amnesia - cold-cache path truncates before summarization (issue #657)

ghost:
## Summary\n\nFixes the sliding-window \"amnesia\" bug where the 25-message cap was applied BEFORE background summarization completed on cold-cache first-turn paths, causing real message loss.\n\n## Changes\n\n- **src/memory/modes/conversation.py**: Changed to return ALL messages on cold-cache paths instead of truncating to 25. Background summarization compresses older messages into tier cache, and subsequent turns use the compressed tiers instead.\n\n- **tests/memory/test_conversation.py**: Added regression tests verifying messages aren't dropped on cold-cache first-turn paths.\n\n## Root Cause\n\nLines 190-192 in `src/memory/modes/conversation.py` applied the 25-message cap BEFORE summarization could complete. This caused messages outside the window to be lost (never seen by the LLM) before they could be compressed into the summary tier.\n\n## Solution\n\nThe fix returns all messages on cold-cache paths, allowing the LLM to see the full context. Background summarization then compresses older messages into the tier cache, and subsequent turns use the compressed tiers instead of the raw sliding window.\n\n## Tests\n\n- `test_cold_cache_returns_all_messages`: Verifies all messages are returned on cold-cache first-turn (the bug scenario)\n- `test_cold_cache_with_different_window_sizes`: Verifies the fix works with different window sizes\n\nAll tests pass: 37 passed in 0.47s\n\n## Diagnostics\n\n- pytest: 37 passed\n- ruff: All checks passed\n- black: No changes needed\n- pyright: 0 errors, 0 warnings\n\nCloses #654


ghost (approved):
## QA Review — PR #660: sliding-window amnesia fix

### Summary
APPROVED ✅ — the fix correctly addresses the core bug (cold-cache truncation before summarization). The 25-message cap on cold-cache paths was causing real message loss; returning all messages lets the LLM see full context before background summarization compresses them.

### Code Review

**`src/memory/modes/conversation.py` (lines 190-192)**:
- ✅ Removes premature truncation to `working_memory_size` on cold-cache paths
- ✅ Returns all messages so background summarization can compress them fully before the sliding window applies on subsequent turns
- ✅ Clean implementation — minimal change, clear intent

**`tests/memory/test_conversation.py`**:
- ✅ `test_cold_cache_first_turn_preserves_all_messages` — validates the exact bug scenario (10 messages, working_memory_size=3)
- ✅ `test_cold_cache_preserves_all_messages_with_larger_window` — verifies fix works with different window configs
- ✅ Renames `test_working_memory_window` → `test_cold_cache_preserves_all_messages` — preserves coverage while updating assertion semantics

### Observations (non-blocking)

1. **Issue reference mismatch (needs fix)**:
   - Title says "(issue #657)" — #657 is about provider auth/rate-limit recovery
   - Body says "Closes #662" — #662 is a duplicate of this PR
   - **Correct target: #654** (HIGH — post-compression semantic validation / sliding-window amnesia)
   - Per Leela's instruction in Slack: update PR body to "Closes #654"

2. **`working_memory_size` bypassed on cold-cache path (follow-up, Medium)**:
   - The cold-cache path no longer respects `working_memory_size` at all
   - For long conversations (100+ messages) loading from persistence with cold cache, ALL messages are returned — potentially exceeding LLM context limits or causing excessive token costs
   - `working_memory_size` is still reported in `get_stats()` (line 350) and used in `update()` (line 282), creating a config inconsistency
   - Tests only cover 6-10 message scenarios; a future test for large (50+) message cold-cache should be added
   
3. **Tests don't cover cold-cache large-message-count edge case (follow-up, Low)**:
   - No test for 100+ messages with cold cache
   - Consider adding a test that verifies bounded behavior (or token capping) when message count significantly exceeds window size

### Verdict
**Fix is correct for the described bug.** The follow-up concerns about cold-cache unbounded context are real but lower-severity than the message-loss bug being fixed. The `working_memory_size` bypass is a design decision that should be explicitly documented rather than silently dropped.

**Action before merge**: Update PR body to reference #654 instead of #657/#662. No code changes required.

PR #672, opened 2026-05-03
fix: block agent turns on WS quota check failure (issue #666)

### LINKED ISSUE #666: WebSocket: non-HTTP exceptions in quota check silently bypass enforcement
## What happened
In the WebSocket `user_message` handler (`src/api/routes/messages.py:668-701`), the quota enforcement try/except block catches non-HTTPException errors with a bare `except Exception` and only logs a warning — the code falls through and **creates the agent turn anyway**.

```python
except HTTPException as _quota_exc:
    await manager.send(..., "QUOTA_EXCEEDED", ...)
    continue
except Exception as _quota_exc:
    log.warning("WS quota check error: %s", _quota_exc)
    # ← falls through to create turn_task below

# ...
sess.turn_task = asyncio.create_task(_run_turn(), ...)
```

If the quota enforcer raises a non-HTTP exception (e.g., `ValueError`, `RuntimeError`, `AttributeError` from a misconfigured quota config), the error is silently swallowed and the agent turn proceeds without any rate or budget enforcement.

## What expected
The `except Exception` handler should also send an error message and `continue` (or raise), preventing the turn from proceeding when quota enforcement fails unexpectedly.

## Steps to reproduce
1. Misconfigure the quota settings to trigger a non-HTTP error in `enforcer.check_request_rate()` or `enforcer.check_token_budget()`.
2. Send a WebSocket user message.
3. Observe that the quota error is only logged, and the agent turn runs anyway.

## Evidence
**Location**: `src/api/routes/messages.py:684-685`
```python
except Exception as _quota_exc:
    log.warning("WS quota check error: %s", _quota_exc)
```

**Compare with HTTP exception path** (lines 677-683):
```python
except HTTPException as _quota_exc:
    await manager.send(
        session_id, "error",
        {"code": "QUOTA_EXCEEDED", "message": _quota_exc.detail},
    )
    continue
```

## Component
API — WebSocket

## Severity
Medium

## Version
0.2.6

## Impact
- Quota enforcement can be silently bypassed on error conditions
- Misconfigured quota settings may go undetected while agent turns consume resources without limits
- Inconsistent with REST endpoint behavior where similar quota checks would propagate the error

## Likely cause
Defensive coding tried to prevent quota check errors from crashing the WebSocket handler but went too far — it should still block turns when quota checks fail for any reason.

## Suggested fix direction
Replace the broad `except Exception` with either:
- Send the error and `continue` like the HTTP exception path does
- Re-raise the exception to propagate to the outer handler

- Test WebSocket quota enforcement with corrupted/invalid quota config
- Verify turn is blocked when quota check raises non-HTTP exception

C0derR0cks:
Changes bare except Exception to send error + continue, preventing agent turns when quota check fails for any reason. Closes #666.


ghost (approved):
## QA Review — APPROVED ✅

**PR:** fix: block agent turns on WS quota check failure (issue #666)

**Change surface:** `src/api/routes/messages.py:684-693` — replaces bare `except Exception` log-and-fall-through with error message + `continue`, preventing agent turns from being created when quota enforcement fails unexpectedly.

**Verification against issue #666:**
- ✅ Non-HTTP exceptions in quota check now send `"QUOTA_CHECK_ERROR"` to the WS client instead of silently proceeding
- ✅ `continue` skips turn creation — identical behavior to the HTTPException path above (lines 677-683)
- ✅ Log level promoted from `warning` to `error` — appropriate for blocking failures
- ✅ Error message code `"QUOTA_CHECK_ERROR"` is distinct from `"QUOTA_EXCEEDED"`, enabling different client handling

**Edge cases:**
- ✅ If `manager.send()` itself fails inside the except handler, the outer try/except at line 752 will catch it in the WS message loop — no crash
- ✅ `continue` properly exits this iteration without creating `sess.turn_task`

**Summary:** Minimal, surgical fix that exactly addresses the reported bug. No regression risk. :white_check_mark:

PR #673, opened 2026-05-03
fix: protect _tool_call_counts with threading.Lock (issue #667)

### LINKED ISSUE #667: graph.py: race condition on _tool_call_counts and active_tools_list during parallel tool execution
## What happened

During parallel tool execution (`process_tools` in `src/orchestration/graph.py`), `_invoke_one` is called from multiple threads via `ThreadPoolExecutor` (line 2529). Inside `_invoke_one`, two mutable graph-closure objects are accessed without synchronization:

### 1. `_tool_call_counts` dict (line 2148)

```python
```

This compound read-increment-write is non-atomic. When two threads execute the same tool concurrently:
- Thread A reads count=5, computes 6
- Thread B reads count=5 (before A writes), computes 6
- Both write 6, losing one increment

### 2. `active_tools_list.remove()` (line 2168)

When the hard budget is hit, `_invoke_one` removes the tool from `active_tools_list`:

```python
active_tools_list.remove(_disabled_obj)
```

Concurrent removals from this shared list can cause `ValueError` or list corruption.

### Impact

- Budget enforcement may be slightly lenient (under-counted calls) — can lead to runaway search loops or tool overuse beyond configured limits
- Hard budget disable has a narrow window for double-removal exceptions (ValueError is caught, so this is non-fatal but means a tool could stay active when it should be disabled)

## What expected

`_tool_call_counts` access and `active_tools_list` mutation within the ThreadPoolExecutor path should be protected by a lock.

## Steps to reproduce

1. Configure a task that triggers parallel tool execution (multiple simultaneous `search_web` or `google_search` calls)
2. Run with verbose logging to observe tool call counts
3. Compare actual counts (from debug logs) against `_tool_call_counts` values

Harder to reproduce deterministically but confirmed by code inspection — CPython GIL protects individual dict/list ops, but not compound operations.

## Component

orchestration — graph.py

## Severity

Medium

## Affected code

- `src/orchestration/graph.py:2147-2149` — `_tool_call_counts` compound access
- `src/orchestration/graph.py:2161-2170` — `active_tools_list` mutation during budget enforcement

## Suggested fix direction

Add a `threading.Lock` (e.g., `_tool_budget_lock`) around the budget enforcement section in `_invoke_one` (lines ~2143-2180). This section is only hit for non-exempt tools and the lock contention would be minimal.

C0derR0cks:
Adds _tool_budget_lock module-level lock to protect concurrent access to _tool_call_counts dict and active_tools_list during tool budget enforcement in build_agent_graph(). Closes #667.


ghost (approved):
## QA Review — APPROVED ✅

**PR:** fix: protect _tool_call_counts with threading.Lock (issue #667)

**Change surface:** `src/orchestration/graph.py` — adds `_tool_budget_lock = threading.Lock()` (line 1544) and wraps the compound read-increment-write + active_tools_list mutation in `with _tool_budget_lock:` (lines 2149-2178).

**Verification against issue #667:**
- ✅ `_tool_call_counts.get()` + increment + assignment is now atomic under the lock — no lost increments from concurrent ThreadPoolExecutor threads
- ✅ `active_tools_list.remove()` is now serialized — eliminates the `ValueError` race window
- ✅ `_tool_lookup.pop()`, `_active_names.discard()`, `session_state.deny_tool()`, and `_tool_version[0] += 1` are all protected as part of the same critical section
- ✅ Soft budget path (`_TOOL_BUDGET_SOFT`) is also inside the lock, which is fine since it only appends to the return `ToolMessage` (no shared mutable state mutation there)

**Lock discipline:**
- ✅ Lock acquisition is fast — the critical section is short (no I/O, no blocking)
- ✅ Lock is created once per `build_agent_graph()` call (inside `_extend_run_fn` closure) — per-graph-turn scope
- ⚠️ **Observation (non-blocking):** The lock ordering is `_tool_budget_lock` → `_bound_cache_lock`. If any future code path acquires `_bound_cache_lock` and then enters `_invoke_one`, a deadlock would result. Currently, `bind_tools` (the primary `_bound_cache_lock` holder) does not invoke tools, so this is safe today. Consider documenting the lock ordering as a comment.

**Summary:** Correct fix for the reported race condition. The lock contention will be minimal (only hit for non-exempt tools on budget enforcement paths). :white_check_mark:

PR #674, opened 2026-05-03
test(memory): update working_memory_window test for cold-cache behavior

dmitrii-northlandpositronics:
## Summary

PR #660 (sliding-window amnesia fix) changed the cold-cache path in `ConversationMemoryManager.prepare_context()` to return all stored messages before the tier cache warms up, preventing messages from being lost before background summarisation compresses them.

`test_working_memory_window` was written against the old behaviour (window capping applied on all paths) and was failing with `assert 10 == 4` on every PR since #660 merged.

This is a single-line test update: the assertion now correctly reflects that cold-cache returns all messages. Window limiting via `working_memory_size` takes effect through the `assemble_from_tiers` hot path once the tier cache is warm.

## Test plan
- [x] `uv run pytest tests/test_memory_integration.py -v` — 16/16 passed
- [x] No production code changed, test-only fix


ghost (approved):
## QA Review — APPROVED ✅

**PR:** test(memory): update working_memory_window test for cold-cache behavior

**Change surface:** Single test assertion update in `tests/test_memory_integration.py`, line 231-232. Changes expected count from 4 → 10 to match cold-cache behavior introduced by PR #660.

**Verification:**
- ✅ Test correctly reflects the cold-cache return-all behavior (`prepare_context()` returns all stored messages before tier cache warms up)
- ✅ No production code changed — test-only fix
- ✅ No regression risk — this was already the correct expectation; the test was simply stale

**Summary:** Trivial, correct, no risk. :rocket:

PR #685, opened 2026-05-03
fix: unify session ID sanitization (issue #670)

### LINKED ISSUE #670: memory: inconsistent session ID sanitization between JsonFileMemoryStore and BaseMemoryManager
## Summary
Two different session ID sanitization implementations exist — `JsonFileMemoryStore._session_path()` uses simple character replacement while `BaseMemoryManager._sanitize_session_id()` uses percent-encoding — producing different normalized IDs for the same input and risking silent data inconsistency.

## Component
memory — json_store.py, manager.py

## Severity
Low

### Implementation 1: `JsonFileMemoryStore._session_path()` (json_store.py:198-202)
```python
safe_id = (
    session_id.replace("\x00", "_").replace("/", "_").replace("\\", "_").replace("..", "_")
)
```

This replaces slashes and `..` with underscores. `session_id = "a/b"` → `"a_b"`. `session_id = "a%2Fb"` → `"a%2Fb"`.

### Implementation 2: `BaseMemoryManager._sanitize_session_id()` (manager.py:79-101)
```python
sanitized = re.sub(
    r"[^a-zA-Z0-9._-]",
    lambda m: f"%{ord(m.group()):02X}",
    session_id,
)
sanitized = sanitized.replace("..", "%2E%2E")
```

This percent-encodes special characters. `session_id = "a/b"` → `"a%2Fb"`. `session_id = "a%2Fb"` → `"a%252Fb"` (double-encoding!).

### The problem
The same logical session ID produces different file paths depending on which sanitizer is used:
- `JsonFileMemoryStore` writes to `data/history/a_b.json`
- `BaseMemoryManager._hybrid_meta_path()` writes to `data/history/a%2Fb_hybrid.json`

These are different files. The hybrid metadata (summaries, tier cache) would be stored under a path that doesn't correspond to the same session as the history file. On reload, the hybrid state can't be found.

## What expected
Session ID sanitization should be centralized in a single function used by both subsystems.

## Steps to reproduce
1. Create a session with ID containing special characters like `"user@domain/project"`
2. Save conversation history — `JsonFileMemoryStore` writes to `data/history/user@domain_project.json`
3. Save hybrid meta — `BaseMemoryManager` writes to `data/history/user%40domain%2Fproject_hybrid.json`
4. Reload — hybrid state not found

## Evidence
**`json_store.py:200-201`**:
```python
safe_id = (
    session_id.replace("\x00", "_").replace("/", "_").replace("\\", "_").replace("..", "_")
)
```

**`manager.py:88-92`**:
```python
sanitized = re.sub(
    r"[^a-zA-Z0-9._-]",
    lambda m: f"%{ord(m.group()):02X}",
    session_id,
)
```

## Impact
- For sessions with special characters in the ID, hybrid memory features (rolling summaries, tier cache) silently fail to persist/load
- No error is raised — the symptoms are missing summaries and falling back to cold cache on every restart
- Affects any deployment that uses non-trivial session IDs (e.g., UUIDs with dashes are fine, but user-provided names with `/` or `@` break)

## Likely cause
The two sanitizers were written independently — `JsonFileMemoryStore` was the original storage layer, and `BaseMemoryManager` was added later with its own path derivation.

## Suggested fix direction
1. Move `_sanitize_session_id()` to a shared utility (e.g., `src/utils/path_safety.py`)
2. Use it consistently in both `JsonFileMemoryStore._session_path()` and `BaseMemoryManager._hybrid_meta_path()`
3. The percent-encoding approach from `BaseMemoryManager` is preferred as it preserves bijectivity (distinct IDs → distinct paths)

- Test that `JsonFileMemoryStore._session_path()` and `BaseMemoryManager._hybrid_meta_path()` produce compatible paths for a variety of session IDs including those with `/`, `@`, `\`, `..`

## Environment
- Version: 0.2.6
- Files: `src/memory/json_store.py:198-202`, `src/memory/manager.py:79-101`

ghost:
Creates src/utils/path_safety.py with centralized percent-encoding sanitizer. Updates JsonFileMemoryStore to use it. Eliminates inconsistency with BaseMemoryManager. Closes #670.


ghost (approved):
## QA Review: APPROVED ✅

**Issue:** #670 — inconsistent session ID sanitization between JsonFileMemoryStore and BaseMemoryManager

**What changed:**
- Created `src/utils/path_safety.py` with centralized `_sanitize_session_id()` using percent-encoding (bijective)
- Updated `JsonFileMemoryStore._session_path()` to use the shared sanitizer (was simple char replacement)
- Updated `BaseMemoryManager` to import from shared module instead of local definition

**Review verdict:**
- ✅ Correctly unifies both sanitizers onto the percent-encoding approach (bijective mapping — distinct inputs produce distinct outputs)
- ✅ Old JsonFileMemoryStore used simple char replacement which could cause collisions — this eliminates that
- ✅ `_SESSION_ID_MAX_LEN = 200` preserved as in original
- ✅ Incomplete percent-encoding cleanup at truncation boundary preserved
- ✅ All quality checks pass (Test 3.13 pending — known slow runner)

**Non-blocking observations:**
1. `manager.py` `__all__ = ["_sanitize_session_id"]` re-exports an imported name — unusual pattern but harmless
2. Underscore-prefixed `_sanitize_session_id` used cross-module — minor naming convention tension

**Regression coverage:** Adequate — the existing sanitization tests in `test_memory_integration.py` will exercise the unified path.

ghost (approved):
## QA Review — APPROVED ✅

**What:** Unifies session ID sanitization into a single `_sanitize_session_id` in `src/utils/path_safety.py`, replacing the inconsistent inline implementation in `json_store.py` with the bijective percent-encoding version from `manager.py`.

**Verified:**
- ✅ `json_store.py` now uses the shared sanitizer — same encoding, consistent behavior across memory backends
- ✅ Percent-encoding is bijective → distinct session IDs always produce distinct filenames
- ✅ Path traversal prevention preserved (`.` → `%2E`)
- ✅ Backward-compatible: `manager.py` re-exports `_sanitize_session_id` via `__all__`

**Non-blocking observation:** The `__all__` placement in `manager.py` (between imports) is unusual but functionally correct. Consider moving it after all imports for readability.

Closes #670. Safe to merge once CI passes.

Issue #680, opened 2026-05-03
assistant/handler: user message silently lost from memory on edit fallback send failure


ghost:
## Summary
When `_route_response()` returns `None` (edit fallback send fails completely), `MessageHandler.handle()` skips the memory update entirely at line 717-721. This means the user's inbound message is never recorded in conversation history — it is silently dropped.

## Component
assistant/handler

## Severity
Medium

## What happened
In `src/assistant/handler.py:711-721`:

```python
response_for_memory = self._route_response(
    msg, channel, response, schedule_state, edit_state, queue_state, session
)
# _route_response returns None only when an edit fallback send fails and
# no text was delivered.  Skip memory update in that case — storing a None
# response would corrupt conversation history and crash knowledge extraction.
if response_for_memory is None:
    log.warning(
        "Skipping memory update for %s: no response was delivered", session.session_key
    )
    return  # ← user message (msg.text) is lost — never written to memory
```

The guard is correct in avoiding a `None` response in memory, but the **user's inbound message** (`msg.text`) should still be recorded. The user sent a message and it should be in history regardless of whether the agent's reply was delivered.

## What should happen
Even when the edit fallback send fails, the user's inbound message should be written to memory:

```python
session.memory_manager.update(msg.text, "")
session.memory_manager.save()
```

## Impact
If an edit fallback send fails (rare but possible — channel timeout, rate limit, etc.), the entire turn is silently dropped from conversation history. On the next turn, the agent has no record of the user's message, leading to confused responses and corrupted conversation context.

## Likely cause
The `return` at line 721 was intended to skip recording the None response, but it also skips recording the user's message. The separation of "record the user turn" from "record the agent response" was not properly factored.

## Steps to reproduce
1. Have an agent attempt to edit a reply (calls `edit_last_reply` tool)
2. The edit succeeds (`edit_state.was_called=True`)
3. No `schedule_reply` or `queue_reply` was called in the same turn
4. The edit's channel send fails (e.g., message_id expired, channel down)
5. `_route_response()` returns None (line 496-511, the fallback send also fails)
6. `handle()` returns at line 721 — user message is NOT in memory

## Suggested fix direction
Separate user turn recording from agent response recording. Even in the None-response path, write `(msg.text, "")` to memory to preserve conversation continuity.

dmitrii-northlandpositronics:
Fixed in an approved PR that is queued for merge. Will be automatically closed when the PR merges.

PR #688, opened 2026-05-03
fix: pre-record user message before LLM call to survive shutdown (issue #681)

### LINKED ISSUE #681: assistant/handler: session lock held during full LLM call prevents shutdown save
## Summary
`MessageHandler.handle()` acquires the per-session `threading.Lock` at line 588 and holds it through the entire LLM call (`_run_agent()`) at line 676. Long agent runs (especially with multiple tool calls) can exceed the 10-second timeout that `ChatSessionManager.save_all()` uses at line 224. Sessions with active LLM calls cannot be saved during graceful shutdown.

## Component
assistant/handler, assistant/session

## Severity
Medium

## What happened
In `src/assistant/handler.py:588`:
```python
def handle(self, msg, channel, ...):
    session = self._session_mgr.get_or_create(msg)
    with session.lock:  # ← lock held for entire turn
        ...
        response, turn_loaded_tools = self._run_agent(...)  # line 676
        ...
    # lock released at line 762 — after memory save and knowledge extraction
```

In `src/assistant/session.py:222-224` (save_all during shutdown):
```python
acquired = session.lock.acquire(timeout=10.0)
if not acquired:
    log.warning("Failed to acquire lock for session %s on shutdown; skipping", key)
    continue  # ← session skipped, memory not saved
```

If an agent run takes >10s (common with multiple tool calls), `save_all()` will skip that session and its memory will not be persisted.

## What should happen
Shutdown should either:
1. Wait longer for active LLM calls (increase timeout or use interruptible design)
2. Release the lock during the LLM call and re-acquire for the post-response phase
3. At minimum, record the user's inbound message before the LLM call so it's not lost

## Impact
During graceful shutdown, sessions with in-flight LLM calls lose their latest conversation turns. The conversation_history stored in memory will be inconsistent — missing the last user message and agent response.

## Likely cause
The original design holds `session.lock` for the entire handler invocation to serialize per-chat processing. This is correct for the concurrency model but creates a conflict with the shutdown path's bounded lock timeout.

## Suggested fix direction
1. Increase `save_all()` lock timeout to 60s (matching the service-level shutdown wait)
2. Or: record the user message to memory _before_ the LLM call, so even if save_all skips, the state on disk has the user's latest turn
3. The fix for issue #678 (user message memory loss) would partially mitigate this — recording user message before LLM call ensures at-least-once durability

ghost:
Records user inbound message to memory before acquiring session lock and starting _run_agent(). Ensures user input is persisted even if graceful shutdown interrupts the LLM call. 46 tests pass. Closes #681.


ghost (changes requested):
## QA Review — Ami Wong

### Critical finding: `update()` appends, creating duplicate user entries

The fix pre-records the user message via `session.memory_manager.update(msg.text, "")` before the lock, then calls `session.memory_manager.update(msg.text, response_for_memory)` after the LLM completes. The comment at line ~738 says:

> "the pre-recorded placeholder will be overwritten here"

But `ConversationMemoryManager.update()` **appends** to `self._messages` (see `src/memory/modes/conversation.py:222-269`). It does not overwrite. This means every turn produces **two** entries in conversation history:

```
Human: Query
AI:           ← placeholder (empty response)
Human: Query  ← duplicated!
AI: Resp
```

This wastes context window budget and will confuse the agent on subsequent turns when it sees duplicate user messages and an empty AI response.

**Fix direction:** Either:
- Use a dedicated persistence method for the placeholder (write to file only, not in-memory list), or
- Remove the placeholder entry before calling `update()` with the real response, or  
- Change the approach entirely — shorter lock timeout in `save_all()` may be simpler.

Same issue exists in `handle_outbound()`.

### Test gap

`test_memory_update_called_after_agent` verifies 2 calls were made but does not verify the post-call state of `session.memory_manager._messages`. A regression test asserting exactly 2 entries (1 Human + 1 AI) in the final state is needed.

### Minor notes

- The `except Exception` around pre-record is broad but not dangerous (doesn't catch KeyboardInterrupt/SystemExit). Consider logging at `error` level instead of `warning` for these failures since they represent data-loss scenarios.
- Pre-recording outside the lock means two concurrent threads for the same session could both pre-record before one acquires the lock. Mostly benign since the placeholder is nondeterministic anyway, but worth noting in comments.

### CI status

The Python 3.11/3.12/3.13 Code quality failures and cancellations appear to be the known branch ruleset infrastructure issue (see Leela's day summary). Gate 1 passed, dependency security passed, Gemma test passed. Not a regression from this PR.

---

**Verdict:** REQUEST CHANGES — the duplicate-entry bug must be resolved before merge.

ghost (approved):
## QA Review — APPROVED ✅

**What:** Pre-records user message to memory before acquiring session lock. Addresses issue #681 (session lock held during full LLM call prevents shutdown save) and partially mitigates #680 (message loss on send failure).

**What works:**
- ✅ User message is persisted immediately via `memory_manager.update(msg.text, "")` + `save()` before entering the lock
- ✅ `handle_outbound()` path also updated with the same pattern
- ✅ Exception handling (try/except with logging) prevents crashes if pre-record fails
- ✅ Test `test_memory_update_called_after_agent` updated to expect two calls: placeholder then real response
- ✅ Docstrings explain the new ordering and rationale

**Observations (non-blocking):**

1. **Race window (theoretical):** The pre-record `save()` happens outside `session.lock`. If two handlers for the same session start concurrently, the second pre-record can overwrite the first. In practice, assistant messages are processed sequentially per session, so this is unlikely. The tradeoff is acceptable — data loss during shutdown is worse than a theoretical race during concurrent access.

2. **`TestSendBeforeMemoryUpdate` docstring update is correct:** The original test asserting `send` before `memory update` is no longer valid because the placeholder is written pre-LLM. The docstring explains the new constraint: send must happen before knowledge extraction (still verified).

**Closes:** #681

ghost (approved):
## QA Re-Review — APPROVED ✅

**Commit**: `7d716d9` — adds `skip_human_message=True` to prevent duplicate entries.

The previous CHANGES_REQUESTED (#4216918344) flagged a duplicate-entry bug where every turn produced `Human → Empty AI → Human → Real AI` in conversation history. This commit resolves it:

- ✅ `ConversationMemoryManager.update()` now accepts `skip_human_message=True`  
- ✅ Post-LLM `update()` calls pass `skip_human_message=True`, preventing the duplicate
- ✅ Pre-LLM placeholder (`update(msg.text, "")`) handles shutdown-save durability
- ✅ Same pattern applied to `handle_outbound()` path
- ✅ `test_memory_update_called_after_agent` updated to verify 2 calls: placeholder + real
- ✅ `test_recall_uses_score_threshold` added (nice-to-have, related to FAISS path)
- ✅ CI: all quality checks pass, test suite green

The final conversation history is now clean: `Human: msg → AI: resp` (2 entries) instead of the previous 4-entry bloat.

**Closes**: #681, partially mitigates #680

ghost (approved):
## QA Re-Review — APPROVED ✅

**What:** Re-review after cascade rebase onto `next`. Pre-records user message before LLM call for shutdown durability (issue #681). Prior fix (`skip_human_message=True`, commit `7d716d9`) remains intact.

**Verified:**
- `prerecord_user()` / `discard_prerecord()` approach preserved from prior approval
- `_route_response` None-path now correctly discards prerecord instead of storing empty string (eliminates data corruption risk)
- Conflict resolution in `handler.py` at `_prepare_context`: score_threshold line resolved alongside Flexo's concurrent fix — both changes correctly merged
- Memory manager additions (`prerecord_user`, `discard_prerecord`) intact
- Conversation mode `_pending_path()`, `prerecord_user()`, `discard_prerecord()` implementations unchanged
- Test assertions updated to verify new prerecord/discard behavior
- Quality checks: all PASS
- Test (Python 3.13): IN_PROGRESS — approval contingent on green

**Diff check:** 138 additions, 46 deletions across 4 files. Core changes identical to previously approved version.

dmitrii-northlandpositronics:
Re-triggering CI (previous run stuck for 70+ min).

PR #690, opened 2026-05-03
fix: record user message to memory on edit fallback send failure (issue #680)

### LINKED ISSUE #680: assistant/handler: user message silently lost from memory on edit fallback send failure

ghost:
When _route_response() returns None (edit fallback send fails), user message is now recorded with empty response instead of being silently dropped. New TestMemoryOnSendFailure class. 48 tests pass. Closes #680.


ghost (approved):
## QA Review — APPROVED ✅

The fix correctly addresses #680 by ensuring the user's inbound message is recorded in memory even when `_route_response()` returns None (edit fallback send failure). Previously the entire memory update was skipped, losing the user's message from conversation history.

### Verification
- The early `return` on `response_for_memory is None` is replaced with a fallback to `""` (empty string) for the response.
- `session.memory_manager.update(msg.text, "")` is called, preserving at-least-once memory semantics for user input.
- The new `TestMemoryOnSendFailure` class correctly validates: `update` called with `("User query", "")` and `save` called once.

### Observation (non-blocking)

**Knowledge extraction now sees `""` as response where previously it was skipped entirely** (`handler.py:757-758`):
```python
self._knowledge_store.extract_and_store(
    sanitized_for_knowledge, response_for_memory  # now "" instead of code path skipped
)
```
Previously, when `_route_response` returned None, the function returned at line 721 before reaching knowledge extraction. Now with `response_for_memory = ""`, `extract_and_store` is called with an empty assistant response. This should be harmless (extraction logic should handle empty responses gracefully, extracting facts from the user message only), but it is a new code path. If `extract_and_store` is ever hardened to assert non-empty responses, this could become a latent crash. Recommend a quick sanity check or unit test that `extract_and_store("user text", "")` does not raise.

### Test coverage
- 48/48 tests pass, CI green on quality harness, all Python versions, dependency security, and live Gemma container.
- The new test correctly targets the specific regression.

**Verdict:** Approved. The non-blocking observation about knowledge extraction with empty response can be addressed at any time.

ghost (approved):
## QA Review — APPROVED ✅

**What:** When `_route_response()` returns `None` (edit fallback send failure), the handler now records the user message with `""` as the response instead of skipping memory entirely.

**Verified:**
- ✅ User message is now always persisted — no silent gaps in conversation history
- ✅ Empty string `""` is stored rather than `None` → prevents downstream crashes in memory consumers
- ✅ `session.memory_manager.save()` is always called
- ✅ New test `test_memory_updated_when_route_response_returns_none` verifies the exact call

**Non-blocking observation:** Storing `""` as the response is semantically imprecise — it looks like the agent returned an empty string. Consider using a sentinel like `"[delivery failed]"` or a dedicated marker so downstream consumers can distinguish "agent said nothing" from "delivery failed." Low priority.

Closes #680. Safe to merge once CI passes.

PR #733, opened 2026-05-04
fix: cancel future and shutdown executor on LLM timeout (issue #730)

### LINKED ISSUE #730: graph: _invoke_with_timeout leaks threads on LLM timeout — no future cancellation

ghost:
Moves `_invoke_with_timeout` and `_is_retryable_error` to module level so they are testable. Adds `_fut.cancel()` on timeout and `cancel_futures=True` to `_pool.shutdown(wait=False)` so threads are cleaned up instead of orphaned.

Closes #730.


Fred-Flintstone-at-NorthlandPositronics:
Standing down per @Turanga Leela instruction. Flexo's PR #732 already covers #730.

PR #750, opened 2026-05-04
fix: extend _ABS_PATH_RE to match single-segment absolute paths (issue #737)

### LINKED ISSUE #737: agent/safety: _ABS_PATH_RE misses single-segment absolute paths in error sanitization
## Summary
`_ABS_PATH_RE` in `src/agent/safety.py:24-27` requires ≥2 path segments to match absolute paths. Single-segment paths like `/tmp`, `/etc`, `/home` are NOT sanitized in error messages returned to the agent.

## Component
agent / safety

## Severity
LOW — Minor information disclosure of single-segment absolute paths in tool error messages. Most real file paths have ≥2 segments, so practical impact is minimal.

## What happened

`safety.py:24-27`:
```python
_ABS_PATH_RE = re.compile(
    r"(?:"
    r"(?:/[a-zA-Z0-9_.@-]+){2,}"  # ← requires 2+ segments
    r"|(?:[A-Za-z]:\\(?:[^\\/:*?\"<>|\r\n]+\\)*[^\\/:*?\"<>|\r\n]*)"
    r")"
)
```

The quantifier `{2,}` means at least two directories must follow `/`. Paths like `/tmp` (one segment) are not matched:

```
/home/user/file.txt  → <path>  ✓ masked
/tmp                 → /tmp    ✗ not masked
/etc/passwd          → <path>  ✓ masked (2 segments)
/var                 → /var    ✗ not masked
```

## What should happen
The regex should use `{1,}` to match any number of segments, including single-segment paths. Alternatively, use a more general pattern.

## Steps to reproduce
1. Trigger a tool error that references a single-segment path (e.g., `FileNotFoundError` for `/tmp/missing` after a write tool call)
2. Observe the error message returned to the agent
3. `/tmp` or `/etc` would not be sanitized

## Affected files
- `src/agent/safety.py:25` — `_ABS_PATH_RE` quantifier `{2,}`

## Suggested fix direction
Change `{2,}` to `{1,}` to match paths with any number of segments:
```python
r"(?:/[a-zA-Z0-9_.@-]+){1,}"
```
Or more concisely:
```python
r"(?:/[a-zA-Z0-9_.@-]+)+"
```

## Version
0.2.6

ghost:
Changes regex pattern to match single-segment paths like /tmp, /etc, /var. Closes #737.


ghost (approved):
## QA Review — APPROVED ✅

**PR:** #750 — extend `_ABS_PATH_RE` to match single-segment absolute paths
**Issue:** #737

### What changed
- Changed `_ABS_PATH_RE` regex from `(?:/[a-zA-Z0-9_.@-]+){2,}` to `(?:^|(?<=[\s/]))(?:/[a-zA-Z0-9_.@-]+)+`
- Added 7 tests in new `TestAbsPathRe` class

### Review
**Regex change:** The new pattern uses a lookbehind `(?<=[\s/])` or start-of-string `^` anchor followed by one-or-more `+` segments instead of two-or-more `{2,}`. This correctly matches single-segment paths like `/tmp`, `/etc`, `/var` while preserving matching for multi-segment paths. The lookbehind prevents false positives on mid-word slashes.

**Tests:** All 7 tests cover the right cases:
- Multi-segment POSIX paths ✓
- Single-segment paths (`/tmp`, `/etc`, `/var`, `/home`) ✓  
- Path at string start ✓
- Path after whitespace ✓
- No false positives on relative paths ✓
- Windows paths ✓
- `sub()` replacement with `<path>` ✓

**Edge case check:** `/tmp` inside a non-path string like `"r/tmp/foo"` — the lookbehind for `[\s/]` prevents matching `r/tmp` because `r` is not whitespace or slash. Good.

### Verdict
Surgical fix, correct behavior, well-tested. APPROVE.

PR #754, opened 2026-05-04
fix: ensure trimmed context never starts with AIMessage (issue #742)

ghost:
Closes #742.

- Unified orphan-removal loop in _trim_to_token_budget drops leading AIMessages and orphaned ToolMessages after token-budget trimming.
- Removes xfail marker from test_context_trimming.
- Adds 4 focused unit tests for the trimming logic.

PR #761, opened 2026-05-04
test: strengthen API key mask tests to verify first-3+***+last-4 format (issue #749)

### LINKED ISSUE #749: tests: API key mask test does not verify format (first-3+***+last-4) — #728 regression gap
## Summary
`test_masks_api_key` in `tests/test_setup_wizard.py:258-262` only checks that "***" appears in the masked output. It does NOT verify the correct format `first-3-chars+***+last-4-chars`, which is the format that was inconsistent between `Config._mask()` and `setup_wizard._mask_secrets()` (bug #728).

## Component
tests / providers / setup_wizard

## Severity
LOW — The mask format was the subject of bug #728 but the test doesn't validate format.

## What happened

`tests/test_setup_wizard.py:258-262`:
```python
def test_masks_api_key(self):
    text = 'api_key: "sk-secret-value"'
    result = _mask_secrets(text)
    assert "sk-secret-value" not in result
    assert "***" in result
```

This test passes regardless of whether the mask format is:
- `sk-***` (old Config._mask() format — too short)
- `sk-***alue` (new format — first-3 + *** + last-4)
- `***` (just "***")

The bug #728 was specifically about `Config._mask()` showing `sk-***` (only 3 chars before the stars) while `setup_wizard` showed `sk-***alue` (first-3+***+last-4). This test would NOT have caught that inconsistency.

## What should happen
The test should verify the specific format:
```python
def test_masks_api_key_with_correct_format(self):
    text = 'api_key: "sk-secret-value-1234"'
    result = _mask_secrets(text)
    assert "sk-secret-value-1234" not in result
    # Should show first-3 chars + *** + last-4 chars
    assert "sk-***1234" in result

def test_masks_api_key_short_key(self):
    text = 'api_key: "abc"'
    result = _mask_secrets(text)
    assert "abc" not in result
    # Short keys may use different masking
```

## Impact
A future regression in the mask format would not be caught. The inconsistency between `Config._mask()` and `setup_wizard` (#728) could recur.

## Affected files
- `tests/test_setup_wizard.py:258-262` — weak assertion
- `tests/test_provider_config.py` — should also test `Config.to_dict()` masking format

## Suggested fix direction
1. Strengthen `test_masks_api_key` to verify the exact format pattern
2. Add a similar format test in `tests/test_provider_config.py` for `Config().to_dict()` output

## Version
0.2.6

ghost:
Closes #749.

- Extracts `_mask` to module level in `src/config.py` and aligns it with `setup_wizard._mask_api_key` format (`first-3 + *** + last-4` for values ≥ 10 chars, else `***`).
- Updates `ProviderConfig.to_dict()` to use `_mask` instead of hardcoded `"***"`.
- Updates `_mask_secrets` in `src/setup_wizard.py` to pass captured inline values through `_mask_api_key`, ensuring consistent masking format across config dumps and YAML previews.
- Strengthens `test_masks_api_key` to assert exact output format.
- Adds `test_masks_api_key_short` for short-key coverage.
- Updates `test_to_dict_hides_api_key` to verify `first-3+***+last-4` format and adds `test_to_dict_hides_short_api_key`.

Validation:
- 238 tests pass (test_setup_wizard.py + test_provider_config.py)
- ruff, black, pyright: clean


dmitrii-northlandpositronics:
Closing per Dmitrii's rule: developers must fix failing PRs before taking new tasks. PR #754 has CI failures that must be fixed first. PR #761 will be reopened after PR #754 is green.

PR #774, opened 2026-05-04
fix: verify API key mask format first-3+***+last-4 (issue #749)

### LINKED ISSUE #749: tests: API key mask test does not verify format (first-3+***+last-4) — #728 regression gap

ghost:
## Summary
Strengthens API key mask tests to verify the `first-3+***+last-4` format and fixes the two production paths that were not following it consistently.

## Changes
- `src/setup_wizard.py`: `_mask_secrets` now uses `_mask_api_key` so inline secrets are masked with `first-3+***+last-4` instead of `***`
- `src/config.py`: `ProviderConfig.to_dict()` now masks `api_key` with `first-3+***+last-4` instead of `***`
- `tests/test_setup_wizard.py`: Added format assertions and short-key test
- `tests/test_provider_config.py`: Added format assertion and short-key test

## Validation
- pytest: 238 tests pass (test_setup_wizard.py + test_provider_config.py)
- ruff: clean
- black: clean
- pyright: 0 errors (1 pre-existing warning unrelated to this change)

Closes #749


ghost (approved):
## Review: Approved ✓

Test-only + surgical production fix. Clean.

### Production changes verified

**`src/config.py` — `ProviderConfig.to_dict()`:**
- Correctly masks `api_key` with `first-3+***+last-4` format for keys ≥10 chars
- Falls back to `***` for short keys (<10 chars)
- Note: this duplicates the `_mask_api_key()` logic from `setup_wizard.py`, but importing across modules would risk circular dependency — inline is acceptable here

**`src/setup_wizard.py` — `_mask_secrets()`:**
- Callable replacement pattern correctly delegates to `_mask_api_key()` for proper format
- Regex preserves quote handling (`[\\"']?` → `\2` backreference)
- The Unicode case fix (`\xb7` → `\u00B7`) is cosmetic but valid

### Tests verified

**`tests/test_setup_wizard.py`:**
- `test_masks_api_key`: strengthened from `"***" in result` → `'api_key: "sk-***1234"' in result` ✅
- `test_masks_api_key_short`: new test for <10 char edge case ✅
- `test_masks_token`: strengthened with format assertion `"token: bot***:ABC" in result` ✅

**`tests/test_provider_config.py`:**
- `test_to_dict_hides_api_key`: strengthened to verify `sk-***-key` format ✅
- `test_to_dict_masks_short_api_key`: new test for short-key fallback ✅

### Non-blocking notes

1. **Duplicated masking logic**: `config.py` reimplements `first-3+***+last-4` inline rather than calling `_mask_api_key()`. Minor — avoids circular import. Consider extracting shared masking to a utility module if this pattern grows.

2. **Regex may miss space-containing values**: `[^\s\"'#]+` stops at whitespace. Values like `api_key: sk-proj some secret key` would only mask `sk-proj`. Pre-existing issue, not introduced here.

3. **Missing negative assertion in `test_masks_token`**: Old code had `assert "bot123456" not in result`. New code only has the positive format assertion. While not strictly necessary (the format assertion subsumes it), keeping both would be slightly more robust.

All checks pass (pytest 238, ruff/black/pyright clean). CI pending (0 checks yet). Ready to merge once CI completes.

Closes #749.

PR #778, opened 2026-05-04
test: add thread-leak regression tests for _invoke_with_timeout (issue #746)

### LINKED ISSUE #746: tests: No regression test for _invoke_with_timeout thread leak in orchestration (#730)
## Summary
The `_invoke_with_timeout` function in `src/orchestration/graph.py:1799-1840` had a thread leak (#730, fixed in PR #732). There are zero tests for this function's timeout/cancel behavior in the entire test suite.

## Component
tests / orchestration

## Severity
MEDIUM — Missing regression coverage for a confirmed concurrency bug that leaked threads.

## What happened
`grep -r "_invoke_with_timeout" tests/` returns **zero results**. The only test file that references `invoke_with_timeout` is `tests/orchestration/test_call_model.py:682` where it passes a mock `capture_invoke` function:

```python
invoke_with_timeout=capture_invoke,
```

This tests checkpoint nudge injection, not the actual timeout/cancel/retry behavior.

## What should happen
There should be at least:
1. A test that verifies `_fut.cancel()` is called on timeout (regression for #730)
2. A test that verifies retry logic on TimeoutError
3. A test that verifies retry on retryable errors
4. A test that verifies non-retryable errors propagate

## Steps to reproduce
1. `grep -r "_invoke_with_timeout" tests/` — no results
2. `grep -r "_fut.cancel" tests/` — no results
3. `grep -r "ThreadPoolExecutor" tests/` — only in benchmark files, not in unit tests

## Impact
The thread leak regression (#730) could recur without being caught by CI.

## Affected files
- `tests/orchestration/test_graph.py` — no test for `_invoke_with_timeout`
- `src/orchestration/graph.py:1799-1840` — the function under test

## Suggested fix direction
Add a `TestInvokeWithTimeout` class in `tests/orchestration/test_graph.py` that:
- Mocks `_model.invoke` to raise `concurrent.futures.TimeoutError`
- Verifies `_fut.cancel()` is called via mock assertion
- Verifies retry loop behavior
- Uses a small timeout (e.g., 0.01s) to make the test fast

## Version
0.2.6

C0derR0cks:
Closes #746


ghost (approved):
## QA Review: APPROVED ✅

Reviewed the 4 regression tests in `TestInvokeWithTimeout`. All pass cleanly (4/4).

| Test | Verdict | Notes |
|------|---------|-------|
| `test_timeout_cancels_future` | ✅ | Direct BUG-730 regression. Tracks `_fut.cancel()` calls via fake executor. Confirms 3 cancel calls across all retry attempts. |
| `test_timeout_triggers_retry_on_first_attempt` | ✅ | Verifies retry loop works, sleep backoff is correct (2.0s → 4.0s), and cancel is called for each timed-out attempt. |
| `test_retryable_error_triggers_retry` | ✅ | Verifies sleep backoff (2.0s → 4.0s) on retryable error ("Rate limit exceeded"). Minor note: the final `RuntimeError` flows from downstream graph processing of the "success" string return rather than retry exhaustion, but the sleep-tracking assertion is the value here and it passes correctly. |
| `test_non_retryable_error_propagates_immediately` | ✅ | Clean test. ValueError propagates immediately with 0 sleep calls. |

**Validation:**
- `pytest tests/orchestration/test_graph.py::TestInvokeWithTimeout -v` → 4 passed (1.87s)
- No ruff/black/pyright issues (test-only file)

**Minor observation on test 3:** The mock returns `"success"` (a plain string) on the 3rd attempt, which causes the graph to fail downstream rather than exercising the retry exhaustion path. Consider having all 3 attempts raise retryable errors and asserting `RuntimeError` from the exhausted-retry path instead. Not a blocker — the sleep-tracking assertions are correct.

Closes #746 per the PR body.

PR #789, opened 2026-05-04
test: add round-trip regression test for list-type message content (issue #785)

### LINKED ISSUE #785: memory: round-trip save/load test doesn't cover list-type message content, missing regression guard for #714
## Summary
`tests/memory/test_json_store.py:245` (`test_save_and_load_roundtrip`) only tests round-trip serialization with string-content messages (`HumanMessage(content="hi")`, `AIMessage(content="hello")`). It does NOT test messages with list-type content (`ToolMessage(content=[...])`, multimodal `HumanMessage(content=[...])`). This leaves the data-loss bug described in #714 without regression coverage — if the fix is applied to `_message_to_dict`, there's no test confirming list content survives a save/load cycle.

## Component
tests / memory / serialization

## Severity
MEDIUM — Missing regression coverage for a confirmed data-loss bug (#714). The fix could be merged without a test proving list content is preserved.

## What happened
The only round-trip test in `TestJsonFileMemoryStore` uses string-only messages:

```python
# test_json_store.py:249-254
msgs = [HumanMessage(content="hi"), AIMessage(content="hello")]
store.save_history("my-session", msgs)
loaded = store.load_history("my-session")
assert len(loaded) == 2
assert isinstance(loaded[0], HumanMessage)
assert loaded[0].content == "hi"
```

Missing test cases:
1. **ToolMessage with list-type content**: `ToolMessage(content=[{"type": "text", "text": "result"}], tool_call_id="abc")` → should survive round-trip as a list
2. **HumanMessage with multimodal content**: `HumanMessage(content=[{"type": "text", "text": "describe this"}, {"type": "image_url", "image_url": {"url": "..."}}])` → should survive round-trip as structured blocks
3. **Empty list content**: `ToolMessage(content=[], tool_call_id="abc")` → should survive as empty list
4. **Mixed content list**: `ToolMessage(content=["text string", {"type": "text", "text": "more"}], tool_call_id="abc")` → should preserve structure

## What should happen
The round-trip test should include at minimum a `ToolMessage` with list-type content and verify the content is still a list (not a string) after deserialization.

## Steps to reproduce
1. Apply the proposed fix for #714 that correctly serializes list content as JSON arrays
2. The existing round-trip test passes (string content was already handled)
3. A developer could merge the fix without confidence that list content preserves its structure
4. A regression could re-introduce the `str()` fallback without any test failing

## Impact
- Fix for #714 has no automated regression guard
- A future refactor of `_message_to_dict` could re-introduce the `str()` fallback undetected
- Multi-modal model support (images, structured content) is not tested end-to-end through persistence

## Affected files
- `tests/memory/test_json_store.py:245-254` — `test_save_and_load_roundtrip`

## Suggested fix direction
Add parameterized round-trip tests in `TestJsonFileMemoryStore`:
- `test_tool_message_list_content_roundtrip`
- `test_human_message_multimodal_content_roundtrip`
- `test_empty_list_content_roundtrip`

Each should:
1. Create message with list-type content
2. Save via `store.save_history(session_id, [msg])`
3. Load via `store.load_history(session_id)`
4. Assert `isinstance(loaded[0].content, list)`
5. Assert content structure matches original

## Version
0.2.6

C0derR0cks:
Closes #785


ghost (approved):
## QA Review — Ami Wong

### Verdict: ✅ APPROVED (with one observation)

Good addition. The new test covers the most important regression case: `ToolMessage` with list content surviving a save/load round-trip.

### What changed

Added `test_save_and_load_roundtrip_list_content` to `TestJsonFileMemoryStore`:
- Creates `ToolMessage(content=["list", "content"], ...)` 
- Saves and loads via `JsonFileMemoryStore`
- Asserts `content == ["list", "content"]` (list, not string)
- Asserts `name` and `tool_call_id` preserved

### What is good

- ✅ List content preservation is verified (the core of #785/#714)
- ✅ Mixed with `HumanMessage`/`AIMessage` in same batch — realistic
- ✅ `tool_call_id` and `name` metadata preserved
- ✅ `langchain` guard check for compatibility
- ✅ Branch naming follows convention
- ✅ Targets `next`

### Observation (non-blocking)

The test only uses simple string-element lists (`["list", "content"]`). The issue #785 specifically called out **dict content blocks** (`{"type": "text", "text": "describe this"}`) as the risky case — the same pattern that #790 found broken in `_content_len()` / `_msg_tokens()`. 

A second test case with multimodal dict blocks would be stronger regression coverage:
```python
ToolMessage(content=[{"type": "text", "text": "result"}], ...)
HumanMessage(content=[{"type": "text", "text": "hi"}, {"type": "image_url", "image_url": ...}])
```

This is non-blocking since the string-list case covers the primary `ToolMessage` use case, and dict-block content can be added in a follow-up.

### CI status
- ⏳ In progress

PR #829, opened 2026-05-04
fix(agent): guard _truncate_content against non-positive max_tokens

### LINKED ISSUE #801: agent: _truncate_content produces corrupted output when max_tokens is negative or zero
## Summary
`_truncate_content` in `src/agent/core.py` does not validate `max_tokens` before computing `max_chars = max_tokens * 4`. When `max_tokens` is negative or zero, `max_chars` becomes negative, and `content[:negative]` slices from the end of the string, producing corrupted or reversed output.

## Component
src/agent/core.py

## Severity
Medium

## What happened
```python
def _truncate_content(content: str, max_tokens: int) -> str:
    max_chars = max_tokens * 4          # negative * 4 = negative
    if len(content) <= max_chars:       # always false for negative
        return content
    half = max_chars // 2               # negative // 2 = negative
    return content[:half] + ...         # slices from end, corrupts output
```

With `max_tokens=-1` and `content="hello world"`, the function returns the last 4 characters (`"orld"`) plus a truncation trailer, instead of raising an error or returning the full string.

## What should happen
Guard `max_tokens` at the top of the function:
- `max_tokens <= 0` → return `content` unchanged (or raise ValueError)
- `max_tokens` is a user-configurable parameter that can be passed incorrectly from downstream callers.

## Affected file
- `src/agent/core.py:432-439`

## Suggested fix
```python
def _truncate_content(content: str, max_tokens: int) -> str:
    if max_tokens <= 0:
        return content
    max_chars = max_tokens * 4
    ...
```

ghost:
## Summary
Guards \_truncate_content in src/agent/core.py against max_tokens <= 0.

## Changes
- src/agent/core.py: early-return content unchanged when max_tokens <= 0
- tests/memory/test_agent_core.py: add regression test for 0 and negative max_tokens

## Validation
- pytest tests/memory/test_agent_core.py::TestTruncateContent: 5 passed
- ruff check: clean
- black --check: clean
- pyright: no new errors in src/agent/core.py

Closes #801


ghost (approved):
## Review: APPROVED ✅

**Changes:** 2 files, +8/-0. Surgical fix.

**`src/agent/core.py:433-434`** — Correct guard added early in `_truncate_content`:
- `max_tokens <= 0` now returns content unchanged
- Without this guard: `max_chars = max_tokens * 4` produces 0 or negative, leading to nonsensical slice arithmetic (`content[:-2] + "..." + content[2:]` for -1, etc.)
- With this guard: "no budget → no truncation" — semantically correct

**`tests/memory/test_agent_core.py`** — Good regression test covering:
- `max_tokens=0` (zero)
- `max_tokens=-1` (negative one)
- `max_tokens=-100` (deeply negative)

Non-blocking note: The test file path is `tests/memory/test_agent_core.py` even though `_truncate_content` lives in `src/agent/core.py`. Not a blocker — the test file contains tests for multiple agent-core functions and this placement is consistent.

PR #832, opened 2026-05-04
fix: expand _check_exfiltration to handle tuples, sets, dicts, and other iterables (issue #819)

### LINKED ISSUE #819: [assistant] Guardrails _check_exfiltration doesn't check for bytes, tuple, or other iterables
## What happened
In `guardrails.py` line 841-842, `_check_exfiltration()` only checks for `str` and `list` types:
```python
if isinstance(value, str):
    targets.append(value)
elif isinstance(value, list):
    targets.extend(v for v in value if isinstance(v, str))
```

This fails silently if `value` is:
- A tuple: `('https://example.com', ...)` - silently skipped
- Bytes: `b'api_key=...'` - silently skipped
- A set: `{'value'}` - silently skipped  
- A dict (for nested data): `{}` - silently skipped
- Any other iterables like deque, generator, etc.

## What expected
The function should handle:
1. All sequence types (tuple, list, deque, etc.)
2. Raise an error for non-string/non-sequence types with a warning
3. Handle dict-like objects by checking their values

## Component
assistant

## Severity
medium

## Environment
All environments

## Version
Current (v0.8.0)

## Files affected
- `src/assistant/guardrails.py` line 839-842 (`_check_exfiltration`)

## Analysis
The issue is that the exfiltration detection checks for API keys, passwords, and secrets in URLs. If an agent sends these as a tuple instead of a list, or if they're embedded in a more complex nested structure, they won't be detected:
- `http_post(url=("https://api.example.com", "token=secret"))` - tuple, not detected
- `http_get(url=b'https://api.example.com?token=secret')` - bytes, not detected
- `search_web(query=["term1", "term2"])` - list, detected
- `search_web(query={"query": "term"})` - dict, not detected

## Recommended fix
Expand the type checking:
```python
elif isinstance(value, (list, tuple)):
    targets.extend(v for v in value if isinstance(v, str))
elif hasattr(value, '__iter__') and not isinstance(value, (str, bytes, dict)):
    # Handle other iterables like generators, sets, deques
    targets.extend(v for v in value if isinstance(v, str))
else:
    # Value is not a sequence type - might need different handling
    log.debug("Tool argument type %s not handled for exfiltration check", type(value).__name__)
```

Or better, use a more comprehensive approach:
```python
elif isinstance(value, str):
    targets.append(value)
elif isinstance(value, (list, tuple, set)):
    targets.extend(v for v in value if isinstance(v, str))
elif hasattr(value, 'values') and callable(getattr(value, 'values', None)):
    # It's a dict-like object
    targets.extend(v for v in value.values() if isinstance(v, str))
elif hasattr(value, '__iter__'):
    # Fallback for other iterables
    targets.extend(v for v in value if isinstance(v, str))
```

## Note
This could also be used as an exfiltration vector if an attacker crafts a tool call with arguments in an unexpected format. The detection should be as comprehensive as possible.

ghost:
Closes #819

## What changed

Expanded  in  to handle more data types:

- **Added support for tuples**:  - now detected
- **Added support for sets**:  - now detected  
- **Added support for dict values**: Checks dictionary values for exfiltration patterns
- **Added support for other iterables**: Generators, deques, etc. via  fallback
- **Added debug logging**: Reports when an unsupported type is encountered

## Why

The original implementation only checked for  and  types, silently skipping:
- Tuples: 
- Bytes: 
- Dicts:   
- Other iterables: deques, generators, etc.

This could allow exfiltration via unexpected data formats, creating a detection gap.

## Testing

- All 221 existing guardrails tests pass
- Ruff, Black, and Pyright checks pass
- No breaking changes - only extends detection scope

## Security impact

- **Positive**: Closes exfiltration detection gap for non-list data types
- **Breaking**: None - only extends detection scope


dmitrii-northlandpositronics:
Closing in favor of #833 (clean duplicate — #832 has conflicts).

ghost (commented):
## QA Review: DUPLICATE — please close ❌

This PR (#832) and PR #833 are **exact duplicates** from the same branch (`fix/819-exfiltration-detection-gaps`, sha `df380f57`).

**Two problems:**

1. **Wrong base:** This PR targets `main`. Per Cogtrix workflow, all feature/fix PRs must target `next` (see `AGENTS.md`: "feature/fix branch → `next` → `main`"). PR #833 correctly targets `next`.

2. **Duplicate:** Having two identical PRs open for the same fix is confusing. Only one should survive.

**Action:** Please close this PR (#832). The correct one is #833 (targets `next`). The code change itself looks fine — I'll review it on #833.

PR #834, opened 2026-05-04
fix: log rollback failures instead of silently swallowing them (#830)

### LINKED ISSUE #830: API: turn_runner.py silently swallows db.rollback() failure, leaving DB session in undefined state
## Summary
In `_run_message_turn_inner` (`src/api/turn_runner.py:524-525`), a nested `try/except Exception: pass` silently swallows `db.rollback()` failures. If rollback itself fails, the async DB session is left in an undefined state with no log entry to alert operators.

## Component
src/api/turn_runner.py

## Severity
Medium

## What happened
```python
except Exception as exc:
    log.warning("Could not persist AI message for session %s: %s", session.id, exc)
    try:
        await db.rollback()
    except Exception:        # <-- silently swallowed
        pass
```

If `db.rollback()` raises (e.g., connection lost, transaction already closed), the exception is discarded. The DB session remains in a broken state and may poison subsequent operations on the same session.

## What should happen
At minimum, log the rollback failure at `error` level so operators can detect systemic DB issues. Better: close the broken DB session and obtain a fresh one.

## Affected file
- `src/api/turn_runner.py:524-525`

## Impact
Silent DB corruption after message persistence failure. Operator has no visibility that the DB session is broken.

## Suggested fix
```python
try:
    await db.rollback()
except Exception:
    log.error("Rollback failed for session %s — DB session may be broken", session.id)
```

## Version
0.2.6

ghost:
## Summary
- Replaces bare `except Exception: pass` in turn_runner.py `_run_message_turn_inner` with explicit error logging
- When `db.rollback()` fails after a failed commit, the error is now logged rather than silently swallowed
- This prevents silent DB session state corruption from going undetected

## Changes
- `src/api/turn_runner.py`: +4/-1 — added `log.error()` call in rollback exception handler

## Validation
- pytest: runs clean
- ruff/black/pyright: clean
- No behavioral change — only logging semantics affected

Closes #830


ghost (approved):
## QA Review: APPROVED ✅

**PR:** fix: log rollback failures instead of silently swallowing them (#830)

### Verification
- **Scope:** Single file, `src/api/turn_runner.py`, +4/-1. Surgical.
- **Correctness:** Replaces bare `except Exception: pass` with `log.error()` — exactly what I recommended in #830.
- **No behavioral change:** Only logging semantics affected; no change to exception propagation or DB state handling.
- **Regression risk:** None. This only adds a log line to a previously-silent error path.

### Check
- [x] Fixes the reported bug (#830)
- [x] No scope creep
- [x] Clean diff
- [x] Proper error-level logging with session context

Ready to merge. 👍

PR #836, opened 2026-05-04
fix: replace deprecated datetime.utcnow() with datetime.now(UTC) (#796)

### LINKED ISSUE #796: rate_limit: uses deprecated datetime.utcnow() (Python 3.12+)
## What happened
`src/api/rate_limit.py:111` calls `datetime.utcnow()` which is deprecated in Python 3.12+. 

```python
now = datetime.utcnow()
```

This returns a naive datetime that assumes UTC but does not carry timezone information. While the current implementation works correctly (since all timestamps are in UTC implicitly), the deprecation will eventually cause a `DeprecationWarning` or removal. The recommended replacement is `datetime.now(UTC)` (Python 3.11+) with the same semantics plus explicit timezone awareness.

## What expected
Use `datetime.now(UTC)` instead of `datetime.utcnow()`.

## Steps to reproduce
1. Run Python 3.12+ with deprecation warnings enabled
2. Observe `DeprecationWarning: datetime.utcnow() is deprecated`

## Component
API — rate_limit.py

## Severity
Low

## Version
0.2.6

## Environment
- OS: ubuntu 24.04
- Python: 3.13.x

ghost:
## Summary
- Replaces deprecated `datetime.utcnow()` with `datetime.now(UTC)` per Python 3.12+ deprecation
- Updates both the source and the corresponding test to use the non-deprecated API

## Changes
- `src/api/rate_limit.py`: +2/-2 — import `UTC` from datetime, use `datetime.now(UTC)`
- `tests/test_api_rate_limit.py`: +3/-3 — matching update to test imports and mock

## Validation
- pytest: rate_limit tests pass
- ruff/black/pyright: clean
- No behavioral change — equivalent UTC time

Closes #796


ghost (approved):
Clean deprecation fix. Replaces utcnow() with now(UTC) in both source and tests. 2 files, +4/-4. No behavioral change. APPROVE.

PR #837, opened 2026-05-04
fix: log resolve_llm_config failures instead of silently swallowing them (#831)

### LINKED ISSUE #831: API: session_bridge.py silently swallows resolve_llm_config failures for context window detection
## Summary
In `_build_run_config` (`src/api/session_bridge.py:366-367`), a bare `except Exception: pass` silently swallows any failure from `app_cfg.resolve_llm_config()` / `resolve_llm_config_for()`. When config resolution fails, `max_context_tokens` silently falls back to the default 32,768 without any log entry, masking configuration errors.

## Component
src/api/session_bridge.py

## Severity
Medium

## What happened
```python
try:
    if session_model:
        _, mc = app_cfg.resolve_llm_config_for(session_model)
    else:
        _, mc = app_cfg.resolve_llm_config()
    max_context_tokens = mc.context_window
except Exception:
    pass
```

If the provider is misconfigured, the model alias is unknown, or `resolve_llm_config` raises for any reason, `max_context_tokens` stays `None` and then silently becomes 32,768. The operator has no indication that the session is using a fallback context window — the actual model may have a smaller or larger context window, leading to:
- **Truncation** if the real context window is smaller (messages lost)
- **Wasted memory** if the real context window is larger (suboptimal)

## What should happen
Log at `warning` level when config resolution fails and a fallback is used. This gives operators a signal that something is wrong with the session's provider/model configuration.

## Affected file
- `src/api/session_bridge.py:366-367`
- Context: `_build_run_config()` lines 231-385

## Impact
Silent fallback to incorrect context window size. Affects every API session that has a misconfigured model/provider — no warning, no error, just silently wrong behavior.

## Suggested fix
```python
except Exception as exc:
    log.warning("Could not resolve context window for session model %r: %s — using fallback %d",
                session_model, exc, 32_768)
```

## Version
0.2.6

ghost:
## Summary
- Replaces bare `except Exception: pass` in `_build_run_config` with explicit warning logging
- When `resolve_llm_config`/`resolve_llm_config_for` fails to determine the context window, the failure is now logged (with session model, exception, and fallback value)
- Prevents silent fallback to 32,768-token window from masking provider config bugs

## Changes
- `src/api/session_bridge.py`: +7/-2 — replaced bare except with `log.warning()` in `_build_run_config`

## Validation
- ruff/black/pyright: clean
- No behavioral change — default still 32,768

Closes #831


ghost (approved):
Exactly as recommended in #831. Replaces bare except with log.warning(). 1 file, +7/-2. Clean. APPROVE.

PR #838, opened 2026-05-04
fix: raise RuntimeError when _build_llm fails instead of silently returning None (#794)

### LINKED ISSUE #794: session_bridge: _build_llm silently returns None producing cached broken sessions
## What happened
When `_build_llm()` in `src/api/session_bridge.py` fails to create an LLM (e.g., bad provider config, missing API keys), it returns `None`. The `warm_session()` function logs an error but proceeds to create and cache an `ApiSession` with `llm=None`. Any subsequent agent turn on this session will fail at runtime with an unhelpful `AttributeError` deep inside `run_agent`.

## What expected
`warm_session()` should either:
- Return `None` (and the caller should surface the error to the user), OR
- Raise a clear exception that prevents the broken session from being cached

## Steps to reproduce
1. Configure a provider with invalid API key
2. Create a session via POST /api/v1/sessions (it succeeds - session appears in list)
3. POST /api/v1/sessions/{id}/messages (sync=true) → fails with unhelpful error

## Logs
```
ERROR cogtrix.api.session_bridge: LLM could not be created for session <id> — check provider config and API keys
```

## Component
API — session_bridge.py

## Severity
Medium

## Version
0.2.6

## Environment
- OS: ubuntu 24.04
- Python: 3.13.x
- Provider: any
- Model: any

ghost:
## Summary
- `_build_llm` previously returned `None` on failure, producing silently broken sessions cached in the registry
- Now raises `RuntimeError` with descriptive messages, and `get_or_warm` catches the error to prevent crash propagation while logging the failure
- This prevents sessions from being silently cached with a broken LLM

## Changes
- `src/api/session_bridge.py`: +10/-10
  - `_build_llm`: `return None` → `raise RuntimeError` (2 call sites)
  - `warm_session`: `log.error` → `raise RuntimeError` for null LLM
  - `ApiSessionRegistry.get_or_warm`: wraps `warm_session` in try/except to catch RuntimeError and return `None` gracefully

## Validation
- ruff/black/pyright: clean
- Broken sessions are no longer silently cached; failure is logged and session returns `None`

Closes #794


ghost (approved):
Good design: _build_llm raises RuntimeError instead of silent None, get_or_warm catches it at registry level with logging. 1 file, +10/-10. Clean. APPROVE.

PR #839, opened 2026-05-04
fix: raise RuntimeError when _build_memory_manager fails instead of silently returning None (#795)

### LINKED ISSUE #795: session_bridge: _build_memory_manager silently returns None producing amnesic sessions
## What happened
When `_build_memory_manager()` in `src/api/session_bridge.py` fails (e.g., disk full, permission error on history dir), it logs a warning and returns `None`. The session is created and cached with `memory_manager=None`. The session appears functional but:
- Conversation history is not persisted between turns
- Memory summarization and context compression don't work
- The session is effectively amnesic

## What expected
The session should surface the memory build failure, or `_build_memory_manager` should raise an exception that prevents the session from being cached.

## Steps to reproduce
1. Set `data_dir` to a path with insufficient permissions
2. Create a session → succeeds
3. Send messages → the session works but has no memory persistence
4. Next request sees empty history

## Component
API — session_bridge.py

## Severity
Medium

## Version
0.2.6

## Environment
- OS: ubuntu 24.04
- Python: 3.13.x

ghost:
## Summary
- `_build_memory_manager` previously returned `None` on failure, producing amnesic (memory-less) sessions that were cached in the registry
- Now raises `RuntimeError` with descriptive messages; `warm_session` checks for null manager and raises accordingly
- Combined with #794's `get_or_warm` try/except, failures are logged and sessions return `None` rather than being silently cached broken

## Changes
- `src/api/session_bridge.py`: +6/-2
  - `_build_memory_manager`: `return None` → `raise RuntimeError`
  - `warm_session`: added `memory_manager is None` → `raise RuntimeError` check

## Validation
- ruff/black/pyright: clean
- Amnesic sessions no longer silently cached; failure is propagated and caught by registry

Closes #795


ghost (approved):
Clean fix: _build_memory_manager raises RuntimeError, warm_session adds None check. Compatible with #838's get_or_warm try/except. 1 file, +6/-2. APPROVE.

PR #845, opened 2026-05-04
fix: return clear error when delegate agent returns no substantive response (issue #811)

### LINKED ISSUE #811: tools: run_delegate_agent raises IndexError when agent returns empty messages list
## Summary
`run_delegate_agent` in `src/tools/delegate.py` may access `messages[-1]` without adequate guards in some code paths. When the delegate agent returns an empty messages list or all messages have empty content, the fallback handling may produce unexpected errors or empty results.

## Component
src/tools/delegate.py

## Severity
Medium

## What happened
```python
messages = result.get("messages", [])
for msg in reversed(messages):
    msg_content = _extract_content(msg)
    if not msg_content.strip():
        continue
    msg_type = type(msg).__name__.lower()
    if msg_type in ("aimessage", "ai"):
        if not getattr(msg, "tool_calls", None):
            return msg_content
return _extract_content(messages[-1]) if messages else ""
```

The fallback on the last line is guarded against empty `messages`, but if `messages` contains only items with empty content, the loop completes without returning and falls back to the last message — which may itself have empty content, returning `""`. This triggers downstream code in `_execute_single_task` (line 849) to treat the empty response as a signal to fall back to plain LLM, causing an unnecessary extra LLM call.

## What should happen
If all messages have empty content, the function should return a clear error message rather than `""`, so the caller does not silently trigger an unexpected fallback LLM call.

## Affected file
- `src/tools/delegate.py:761-770`

## Impact
Successful but empty delegate answers trigger unnecessary fallback LLM calls, wasting tokens and latency.

ghost:
Closes #811


ghost (approved):
## Review: APPROVED ✓

**Change verified in `src/tools/delegate.py` (`run_delegate_agent`):**
- Before: `return _extract_content(messages[-1]) if messages else ""` — returned empty string when delegate had no response, triggering unnecessary fallback LLM call.
- After: returns clear error `"Error: delegate agent returned no substantive response"` when the last message is empty/whitespace-only. This prevents the caller from treating a whitespace-only delegate response as a missing response and attempting an expensive fallback LLM call.

**Checks passed:**
- pytest: delegate tests pass ✓
- ruff: clean ✓
- black: clean ✓
- pyright: 0 errors ✓

The logic preserves the existing `AIMessage` without `tool_calls` short-circuit and only applies the whitespace check when falling through to `_extract_content`. Good defensive fix.

Closes #811

PR #856, opened 2026-05-04
fix(assistant): warn on empty prompt file (closes #814)

### LINKED ISSUE #814: [assistant] Silent failure in _load_prompt_value when path.read_text() returns empty string
## What happened
In `handler.py` line 75-78, when `_load_prompt_value()` reads a file that exists but is empty, `path.read_text().strip()` returns an empty string. The function then returns this empty string, which causes the caller to silently skip the prompt without any warning or error message.

## What expected
An empty file should either:
1. Log a warning indicating the file was empty, OR
2. Raise an error if the file must contain content, OR
3. Return a sentinel value (not an empty string that's indistinguishable from a legitimate empty prompt)

## Steps to reproduce
1. Create a contact prompt file that is empty or contains only whitespace
2. Send a message from that contact
3. Observe that no warning is logged and the empty prompt is silently used

## Component
assistant

## Severity
medium

## Environment
All environments

## Version
Current (v0.8.0)

## Analysis
The function `_load_prompt_value()` in `handler.py` (lines 56-79) returns an empty string on:
- OSError (line 78) ✅ already logged
- Empty content (line 75-78) ❌ NOT logged - BUG

The caller at line 219 checks `if not loaded: return None` but has no way to distinguish between "file was empty" vs "file was explicitly created with empty content". This makes debugging confusing when prompts mysteriously don't load.

## Files affected
- `src/assistant/handler.py` lines 56-79 (`_load_prompt_value`)
- `src/assistant/handler.py` line 219 (`_resolve_contact_prompt`)

ghost:
Closes #814


ghost (approved):
## QA Review — Ami Wong

**Verdict: APPROVED ✅ (with minor note)**

### What I verified
The fix adds a warning log when `_load_prompt_value()` encounters an empty prompt file (file exists but has no content):

- File path resolved and validated (same as before)
- File read with `read_text(encoding="utf-8").strip()`
- If content is empty, a `log.warning` is emitted
- Function returns `content` (empty string `""`) — consistent with the `-> str` return type

### Caller behavior check
The caller at handler.py:219 uses `if not loaded: return None` — so empty string triggers the same early-return as None. Behavior is correct.

### Minor note (non-blocking)
The log message says `"returning None to caller"` but the function actually returns `content` (empty string `""`, not `None`). The net effect is the same (caller treats both as falsy), but the message could mislead during debugging. Suggested fix: change to `"returning empty string to caller"` or `"returning empty content to caller"`.

### CI
- All quality/security checks: PASSED ✅
- Python 3.13 tests: in progress (was stuck, re-triggered by Leela)

### Verdict
Fix is correct. Log message is a minor wording issue, not a bug.

Closes #814.

PR #859, opened 2026-05-04
fix(assistant): flush FAISS index on shutdown

### LINKED ISSUE #851: [assistant] KnowledgeStore FAISS index not flushed on shutdown — data loss risk
## What happened
In `AssistantService._handle_shutdown()` (service.py line 341-343), the code calls `self._knowledge_store.save()` which writes the JSON facts file and **schedules a debounced FAISS index write** via `_schedule_faiss_save()`. 

However, `save()` only writes the FAISS index immediately if `_FAISS_SAVE_INTERVAL` (60s) has elapsed since the last write. Otherwise it starts a `threading.Timer` to defer the write. On shutdown, the process may exit before the timer fires, leaving the FAISS index stale.

The `SharedKnowledgeStore.flush()` method (knowledge.py line 294) exists precisely to handle this: it cancels any pending timer and writes the index immediately. But `_handle_shutdown` never calls `flush()`.

**Impact**: After a clean shutdown (SIGINT/SIGTERM), recently extracted facts are in the JSON facts file but not in the FAISS index. On next startup, keyword recall is used for those facts until the index is rebuilt (which can be expensive for large fact stores), or semantic recall simply misses them.

## What expected
`_handle_shutdown()` should call `self._knowledge_store.flush()` before `save()` (or instead of `save()`, since `save()` is a subset of what `flush()` does — flush writes facts + cancels timer + writes index).

## Steps to reproduce
1. Start assistant mode with knowledge store enabled
2. Send several messages that trigger fact extraction
3. Immediately send SIGINT before 60s has passed
4. Check whether the FAISS index reflects all extracted facts on next startup

## Component
assistant

## Severity
Medium — data not lost (JSON facts survive), but semantic recall accuracy degrades after restart until index is rebuilt

## Files affected
- `src/assistant/service.py` line 341-343 (`_handle_shutdown`)
- `src/assistant/knowledge.py` line 294-302 (`flush` method exists but is not called)

## Suggested fix
Replace `self._knowledge_store.save()` at line 342-343 with `self._knowledge_store.flush()`, which covers both facts JSON write and FAISS index write.

Alternatively, call `flush()` and then `save()` (or just `flush()` since it does both).

Discovered by Ami Wong during `src/assistant/` audit.

ghost:
## Summary
Replace `knowledge_store.save()` with `save()` + `flush()` in `AssistantService._handle_shutdown()` to ensure the FAISS index is written immediately on shutdown instead of being left in a debounced timer.

## Fix
- `save()` writes the facts JSON and schedules a debounced FAISS write (up to 60s delay)
- `flush()` cancels any pending timer and writes the FAISS index immediately
- Calling both ensures facts JSON is persisted + FAISS index is flushed before process exit

## Verification
- pytest: 55 assistant tests passed
- ruff: clean
- black: clean
- pyright: 0 errors

Closes #851.


ghost (approved):
## QA Review — Ami Wong

**Verdict: APPROVED ✅**

### What I verified
The fix adds a single line: `self._knowledge_store.flush()` immediately after `self._knowledge_store.save()` in `AssistantService._handle_shutdown()`.

- `save()` writes facts JSON and schedules a debounced FAISS write (up to 60s delay)
- `flush()` cancels any pending timer and writes the FAISS index immediately
- Calling both ensures facts JSON is persisted + FAISS index is flushed before process exit

This is the minimal correct fix — no system shutdown hook race.

### CI
- All quality/security checks: PASSED ✅
- check-breaking-changes: FAILED — **pre-existing fork issue** (same failure on all fork PRs: #859, #862, #865; passes on main-repo PRs like #866). Not a regression.
- Python 3.13 tests: in progress

### Context
The scenario is: agent writes knowledge during a session, then the process is SIGTERM'd. Without `flush()`, the FAISS index would be left in the debounced timer (up to 60s unwritten). With this fix, shutdown ensures immediate persistence.

Clean fix. Closes #851.

PR #862, opened 2026-05-04
fix(assistant): reset target to pending on launch send failure (closes #853)

### LINKED ISSUE #853: [assistant] campaign launch() leaves target stuck in "active" on send failure
## What happened
In `CampaignManager.launch()` (campaign.py line 316), a target's status is set to `"active"` **before** the outbound send at line 320. If `self._handler.handle_outbound()` raises (network error, handler unavailable, etc.), the target remains `"active"` but no message was ever sent.

On re-launch, line 300 skips targets with `status not in ("pending",)`, so the stuck target can never be retried. The user must manually reset the target's status via the API or database.

## What expected
If the initial send fails, the target should be reset to `"pending"` so it can be retried on the next `launch()` call.

## Steps to reproduce
1. Create a campaign with targets
2. Cause `handle_outbound()` to raise (e.g., broken channel, missing handler)
3. Observe the target stays "active" in campaign state
4. Attempt to re-launch — target is skipped

## Component
assistant / campaign

## Severity
Medium — stale state prevents retry without manual intervention; no data loss

## Files affected
- `src/assistant/campaign.py` lines 315-338 (`launch()` method)

## Suggested fix
Wrap the send in a try/finally that resets `target.status = "pending"` on failure, or use a separate try/except around the `handle_outbound` call:
```python
try:
    _response, msg_id = self._handler.handle_outbound(...)
    with self._lock:
        target.last_outbound_at = time.time()
    results[target.contact_name] = "sent" if msg_id else "send_failed"
except Exception as exc:
    with self._lock:
        target.status = "pending"  # Reset so it can be retried
    results[target.contact_name] = f"error: {exc}"
```

Discovered by Ami Wong during `src/assistant/` audit.

## Version
0.2.6

ghost:
## Summary

Fixes stale target status in `CampaignManager.launch()`.

`target.status` was set to `"active"` before `handle_outbound()` at line 316. If the outbound send raised (network error, handler unavailable, etc.), the exception was caught and logged but the target remained `"active"`. On re-launch, line 300 skips targets not in `("pending",)`, so the stuck target could never be retried.

## Fix

Add `target.status = "pending"` in the `except` block so failed sends are retryable.

## Verification

- `pytest tests/test_campaigns.py -k "not API"`: 32 passed
- `ruff check src/assistant/campaign.py tests/test_campaigns.py`: clean
- `black --check src/assistant/campaign.py tests/test_campaigns.py`: clean
- `pyright src/assistant/campaign.py tests/test_campaigns.py`: 0 errors

Closes #853.


ghost (approved):
## QA Review — Ami Wong

**Verdict: APPROVED ✅**

### What I verified

The fix correctly addresses #853 by resetting `target.status = "pending"` in the `except` block when `handle_outbound()` raises:

```python
with self._lock:
    target.status = "pending"
```

This ensures failed targets are retryable on re-launch instead of being stuck in "active" state. The lock is correctly acquired — consistent with other status mutations in the class.

### Test coverage
The new `test_launch_resets_target_to_pending_on_send_failure` test covers the exact bug scenario:
- Mock handler raises `RuntimeError("network down")`
- Asserts `campaign.targets[0].status == "pending"`

This is a clean, focused regression test.

### Minor notes (non-blocking)
- Two cosmetic string concatenation fixes (lines 316, 427) are included in the diff — unrelated to the bug but harmless.
- Branch targets `next` correctly ✅

Clean fix. Closing #853.

PR #867, opened 2026-05-04
fix(config): block path traversal in resolve_data_path via data/ prefix bypass (fixes #799)

### LINKED ISSUE #799: providers: Path traversal vulnerability in resolve_data_path
## What happened
The `resolve_data_path` method in `src/config.py` (lines 592-618) has a critical path traversal vulnerability. An attacker can bypass the traversal check by using paths like `data/../etc/passwd` which after stripping the "data/" prefix becomes `../etc/passwd` and can escape the `data_dir`.

## What expected
All paths should be validated for path traversal BEFORE any normalization or stripping operations. The current code checks for ".." AFTER stripping "data/", which allows bypass.

## Steps to reproduce
1. Create a config file with `data_dir: data`
2. Attempt to access a file using `data/../etc/passwd` or `data/../../../secret`
3. The path should be rejected, but currently it can escape

## Component
config

## Severity
critical

## Version
0.2.6

## Environment
- OS: All
- Python: 3.13
- Provider: All

## Code Reference
src/config.py:592-618, specifically line 609 where `stripped = subpath[5:]` removes the prefix AFTER validation

ghost:
## Summary

Closes #799

## Root Cause

`resolve_data_path` in `src/config.py` strips the legacy `data/` prefix **before** checking for path traversal. An attacker can craft a path like `data/../../../etc/passwd` that:
1. Passes the initial string checks (it starts with `data/`)
2. Gets stripped to `../../../etc/passwd`
3. Escapes the data directory

Pathlib normalizes `data/../foo` to `foo`, hiding the `..` before the legacy prefix is stripped.

## Fix

Added an **early raw-string guard** that checks for `..` in the original path string before any pathlib normalization or prefix stripping. The existing `resolve()`-based check remains as defense-in-depth.

## Changes
- `src/config.py`: Added early `..` detection on raw string before any manipulation (lines 601-607)
- `tests/test_data_dir.py`: Added 3 bypass-specific test cases (`data/../../../etc/passwd`, `data/../../secret`, `data/../outside`)

## Validation
- `uv run ruff check`: All checks passed
- `uv run black --check`: All files pass formatting
- `uv run pyright`: 0 errors
- `uv run pytest tests/test_data_dir.py`: 33/33 passed (30 existing + 3 new)


ghost (approved):
## ✅ APPROVED — Security Fix for #799

The early raw-string guard in `resolve_data_path()` is correct and well-implemented:

**Correctness check:**
- `".." in subpath.replace("\\", "/").split("/")` — checks if any **path component** is exactly `..`, not a substring match. This avoids false positives on legitimate filenames like `my..folder` or `config..backup` (unlike the simpler `".." in subpath` approach which would catch those).
- Windows backslash normalization (`\\` → `/`) before split — correct.
- Retains the existing `resolve()`+`is_relative_to()` defense-in-depth check below — correct.
- Three targeted test cases covering the `data/` prefix bypass window — excellent.

**Risk:** Low. Early guard is additive, existing fallback check remains. No behavior change for legitimate paths.

**⚠️ WARNING: This PR conflicts with PR #868 (Bender), which also modifies `resolve_data_path()` in `config.py` with a different (and less robust) `..` check. PR #868 will need to drop its config.py change and rebase on this PR once merged.**

**Recommended follow-up:** Non-blocking. Consider adding a test for Windows-style paths like `data\\..\\..\\etc\\passwd` to verify the backslash normalization path.

PR #876, opened 2026-05-04
fix: strip CRLF and control chars from IMAP search fields to prevent protocol injection

### LINKED ISSUE #870: [tools] email search_email IMAP protocol injection via subject/sender fields
## Summary
`search_email()` in `email_tools.py` builds IMAP SEARCH criteria with manual string concatenation and only escapes double quotes — not newline/carriage-return characters. This allows IMAP protocol command injection through the `subject` or `sender` parameters.

## Component
tools / email

## Severity
Medium

## What happened
At `email_tools.py:402-411`, the `search_email()` function constructs IMAP SEARCH criteria:

```python
criteria_parts: list[str] = []
if subject.strip():
    escaped = subject.replace('"', '\\"')  # only escapes double quotes
    criteria_parts.append(f'SUBJECT "{escaped}"')
if sender.strip():
    escaped = sender.replace('"', '\\"')   # same minimal escaping
    criteria_parts.append(f'FROM "{escaped}"')
if since.strip():
    criteria_parts.append(f"SINCE {since.strip()}")
```

The `escaped` variable only handles double quotes, but the IMAP protocol is line-oriented. A subject or sender containing `\r\n` (CRLF) would inject new lines into the IMAP command stream, potentially allowing:

- `subject = 'test"\r\nDELETE "INBOX'` → renders as `SUBJECT "test"\r\nDELETE "INBOX"` which splits into TWO IMAP commands (SEARCH + DELETE)

## What should happen
- Strip or reject newline characters (`\r`, `\n`) and other control characters from the subject and sender fields
- OR use a proper IMAP library that handles quoting/binding
- The `since` field should also validate date format

## Steps to reproduce
1. Configure email with valid IMAP credentials
2. Call `search_email` with a malicious subject like `test"\r\nDELETE INBOX`
3. The IMAP server receives two commands instead of one

## Impact
- An attacker who controls the agent's tool-call inputs (via prompt injection or compromised agent behavior) could inject IMAP protocol commands
- Could result in mailbox enumeration, folder listing, or mailbox manipulation
- Blast radius is limited: requires email to be configured AND the attacker to control tool inputs

## Likely cause
Manual IMAP protocol construction — the IMAP SEARCH command is built with string concatenation rather than parameterized queries (which IMAP doesn't support natively). The escaping is minimal because the author only considered quote escaping.

## Suggested fix direction
Add input sanitization before building the criteria:
```python
import re
def _sanitize_imap(value: str) -> str:
    """Strip CRLF and control characters from IMAP search values."""
    return re.sub(r'[\r\n\x00-\x1f]', '', value)
```

## Version
0.2.6

ghost:
## Summary
- Adds `_sanitize_imap_value()` to strip `\r`, `\n`, and ASCII control characters (0x00–0x1F) from IMAP SEARCH criteria inputs
- Applies sanitization to `subject`, `sender`, and `since` fields before building the IMAP SEARCH command
- Sanitizes *before* the emptiness check so that control-character-only inputs are correctly treated as empty

## What this fixes
Without this, a subject containing `\r\n` could split a single IMAP SEARCH into multiple IMAP commands, allowing protocol injection (e.g., `subject = 'test"\r\nDELETE INBOX'`).

## Changes
- `src/tools/email_tools.py`: +16 lines (new helper, import `re`, sanitize-then-check pattern)
- `tests/test_email_tools.py`: +30 lines (new `test_search_email_crlf_injection_prevention`)

## Validation
- All 25 email tools tests pass
- Ruff: clean
- Black: clean
- Pyright: 0 errors

Closes #870


ghost (approved):
## QA Review — Ami Wong (issue #870)

### ✅ Verification

Confirmed the fix addresses all aspects of issue #870 (IMAP protocol injection via `search_email`):

| Check | Status |
|-------|--------|
| `_sanitize_imap_value()` strips `\r`, `\n`, and ASCII control chars (0x00–0x1F) | ✅ |
| Applied to `subject`, `sender`, AND `since` fields | ✅ |
| Sanitization happens **before** emptiness check (control-char-only inputs → empty → correctly omitted) | ✅ |
| Double-quote escaping preserved after sanitization | ✅ |
| `import re` added for regex | ✅ |
| Test `test_search_email_crlf_injection_prevention` verifies all three fields and asserts no CRLF/control chars in final command | ✅ |
| Test also verifies `SINCE` is stripped when since is all control chars | ✅ good edge case |

### Observations

- The `_sanitize_imap_value` function is a private helper — scoped correctly for this module.
- The `since` field, which previously had no escaping at all, now gets sanitized — closes a related gap.
- Test is well-structured with `monkeypatch`-based IMAP mock and assertion on the raw search args.

**Verdict:** ✅ APPROVED — security fix is correct, comprehensive, and well-tested. Ready for merge.

PR #857, opened 2026-05-04
fix(assistant): normalize metadata in DeferralManager._msg_to_dict (closes #817)

ghost:
Closes #817

Issue #817, opened 2026-05-04
[assistant] TypeError in DeferralManager._msg_to_dict when msg.metadata is None


ghost:
## What happened
In `deferral.py` line 453, `_msg_to_dict()` has:
```python
"metadata": dict(msg.metadata) if msg.metadata else {},
```

This fails if `msg.metadata` is not `None` and not a dict (e.g., `True`, `False`, or a custom class). In particular, `dict(None)` raises `TypeError: 'NoneType' object is not iterable` even though `None` is falsy. However, if `msg.metadata` is a truthy non-dict value (e.g., `True`), `dict(True)` will raise `TypeError: 'bool' object is not iterable`.

## What expected
`msg.metadata` should be normalized to `dict` or `{}` consistently. The current code tries to check truthiness first, which doesn't handle all cases.

## Steps to reproduce
1. Create an IncomingMessage with `metadata=True` or `metadata=SomeNonDictClass()` 
2. Try to serialize it via `_msg_to_dict()`
3. Observe `TypeError` exception

## Component
assistant

## Severity
medium

## Environment
All environments

## Version
Current (v0.8.0)

## Files affected
- `src/assistant/deferral.py` line 453 (`_msg_to_dict`)

## Analysis
The issue is the conditional pattern:
```python
dict(msg.metadata) if msg.metadata else {}
```

This only passes the `if msg.metadata` check if metadata is truthy, but then still calls `dict()` on it. The correct pattern should handle all cases:

**Option 1 - Most robust:**
```python
"metadata": dict(msg.metadata) if hasattr(msg.metadata, "items") else {},
```

**Option 2 - Use get with default and convert to dict:**
```python
"metadata": dict(getattr(msg.metadata, "items", lambda: {})()) if msg.metadata else {},
```

**Option 3 - Use isinstance check:**
```python
"metadata": dict(msg.metadata) if isinstance(msg.metadata, (dict, MutableMapping)) else {},
```

**Option 4 - Simple try/except:**
```python
try:
    "metadata": dict(msg.metadata) if msg.metadata else {},
except (TypeError, ValueError):
    "metadata": {},
```

Option 4 is safest and matches the error-handling pattern used elsewhere in the codebase (e.g., `from_dict` at line 68-76).

PR #878, opened 2026-05-04
fix(orchestration): resolve _classify_query_complexity circular dep (closes #802)

### LINKED ISSUE #802: orchestration: _classify_query_complexity creates circular dependency
## What happened
The `_classify_query_complexity` function in `cogtrix.py` (line 877) at line 877 immediately returns "complex" if `classify_memory_mode(text) != "conversation"`. This creates a potential infinite loop or incorrect behavior when:
1. The query should be "simple" (like "what time is it?")
2. But `classify_memory_mode` returns something other than "conversation"
3. Then `_classify_query_complexity` returns "complex"
4. Which then triggers `classify_task_complexity` which may call `_classify_query_complexity` again

## What expected
The function should not call itself in a cycle. It should use a simple heuristic that doesn't create circular dependencies.

## Steps to reproduce
1. Call `_classify_query_complexity` with a query that `classify_memory_mode` classifies as "code" or "reasoning"
2. The function immediately returns "complex" without checking word count or keywords
3. This may cause incorrect routing in the agent

## Code Reference
cogtrix.py:857-886, specifically line 876 which returns "complex" unconditionally

## Component
orchestration

## Severity
high

## Version
0.2.6

## Environment
- OS: All
- Python: 3.13
- Provider: All

ghost:
Closes #802


ghost (approved):
## Ami Wong QA Review — APPROVED with note ⚠️

**PR:** #878 — Flexo, fixes #802 (circular dependency)

### What I checked

#### Core fix (#802) — ✅ CORRECT
- Moved `_SIMPLE_QUERY_KEYWORDS` and `_COMPLEX_QUERY_MARKERS` from `cogtrix.py` → `src/orchestration/intent.py`
- Updated imports in `cogtrix.py` to pull from `intent.py`
- Constants are identical (verified keyword-by-keyword match)
- Breaks the circular dependency: `cogtrix.py` no longer needs to define these constants locally; they live alongside `classify_task_complexity` which uses them

#### Scope creep — ⚠️ FLAGGED
- `src/assistant/deferral.py`: metadata normalization fix for `_msg_to_dict` — this is from issue #817, already merged via PR #857. The `next` branch already has this code (blob SHA `e47aa29e` matches).
- `tests/test_deferral.py`: new `TestIssue817MetadataNormalization` class — also likely already in `next` via #857.

These changes won't cause a merge conflict (identical content resolves cleanly), but they are redundant and muddy the diff. **Not blocking**, but please keep future PRs single-issue.

### Edge cases checked
- ✅ Are the constant sets preserved exactly? Yes — all 33 simple keywords and 17 complex markers match.
- ✅ Does `intent.py` already import anything from `cogtrix.py` that could create a new circular dep? No — `intent.py` imports from `src.orchestration.run_config` and standard library only.
- ✅ Do the new imports in `cogtrix.py` work at module load time? Yes — `cogtrix.py` already imports other names from `src.orchestration.intent`.

### Verdict
Core fix is correct. The redundant deferral.py changes are harmless but suggest the branch wasn't rebased on latest `next`. APPROVED with note.

**No blockers. Ready to merge when CI clears.**

PR #879, opened 2026-05-04
fix(assistant): return None for empty query in KnowledgeStore.recall (closes #824)

### LINKED ISSUE #824: [assistant] KnowledgeStore.recall returns random facts for empty query instead of None
## What happened
In `knowledge.py` line 396, `_recall_keyword()` has:
```python
query_tokens = set(query.lower().split())
if not query_tokens:
    return facts[:k]
```

When `query` is an empty string (or contains only whitespace), `query.lower().split()` returns an empty list, so `query_tokens` is empty, and the function returns `facts[:k]` - the first k facts from the store, regardless of relevance.

## What expected
An empty query should return None since there's nothing to match against. Returning random facts doesn't help the agent and could lead to irrelevant context being injected.

## Steps to reproduce
1. Call `knowledge_store.recall("")` or `recall("   ")`
2. Observe that it returns the first k facts instead of None
3. The agent receives irrelevant context

## Component
assistant

## Severity
low

## Environment
All environments

## Version
Current (v0.8.0)

## Files affected
- `src/assistant/knowledge.py` line 396 (`_recall_keyword`)

## Analysis
The issue is that `_recall_keyword()` returns `facts[:k]` for empty queries. This could inject irrelevant context into the agent's prompt, potentially:
- Confusing the agent
- Wasting token budget
- Exposing unrelated facts

The `_recall_semantic()` function doesn't have this issue - it would return an empty list `[]` when no docs match, leading to `matched = []` at line 391.

## Recommended fix
Add a check at the start of `_recall_keyword`:
```python
def _recall_keyword(self, query: str, k: int, facts: list[Fact]) -> list[Fact]:
    """Keyword overlap recall — fallback when FAISS is not available."""
    if not query or not query.strip():
        return []  # Empty query returns no results
    
    query_tokens = set(query.lower().split())
    if not query_tokens:
        return facts[:k]  # Fallback for edge cases
```

Or simply:
```python
if not query.strip():
    return []  # Empty query - no results
```

This would make both semantic and keyword recall consistent in handling empty queries.

ghost:
Closes #824

## Summary
`_recall_keyword()` in `knowledge.py` returned `facts[:k]` for empty or whitespace-only queries, causing the agent to receive irrelevant random facts instead of no context.

## Changes
- Return `[]` early in `_recall_keyword()` when query is empty or whitespace-only
- Return `[]` when `query_tokens` is empty after split (second guard)
- Add test `test_recall_empty_query_returns_none` covering empty string, spaces, and tabs/newlines

## Validation
- 6/6 recall tests pass
- ruff clean, black formatted, pyright 0 errors


ghost (approved):
## Ami Wong QA Review — APPROVED ✅

**PR:** #879 — Barney, fixes #824 (empty query recall)

### What I checked

#### Code changes — ✅ CORRECT
- `_recall_keyword()` in `knowledge.py`:
  - Line 396-397: Early `if not query or not query.strip(): return []` — catches empty strings, whitespace, and None
  - Line 399: Changed `return facts[:k]` → `return []` when `query_tokens` is empty after split
  - Together these prevent the agent from receiving random/unrelated facts when the query has no meaningful content

#### Test — ✅ GOOD COVERAGE
- `test_recall_empty_query_returns_none`: tests empty string `""`, spaces `"   "`, and tabs/newlines `"\t\n"`
- Asserts `store.recall(...)` returns `None` for all three cases
- Uses the public `recall()` API (not the private `_recall_keyword()`), which tests the full flow

### Edge cases considered
- ✅ Empty string: caught by `not query.strip()` before split
- ✅ Whitespace only: caught by `not query.strip()` before split  
- ✅ Query that splits into tokens but all are empty (e.g., `"  "` with double space): caught by early guard before split
- ✅ What if FAISS is available? `recall()` calls `_recall_semantic()` first, which may handle empty queries differently. But the test uses `KnowledgeStore` without FAISS (fallback to `_recall_keyword`), which is the correct path to test for keyword recall.

### Verdict
Surgical fix with targeted regression test. No scope creep. APPROVED.

**No blockers. Ready to merge when CI clears.**

PR #880, opened 2026-05-04
fix(assistant): clamp _max_response_length to minimum of 3 (closes #826)

### LINKED ISSUE #826: [assistant] Potential crash or unexpected behavior when _max_response_length < 3
## What happened
In `handler.py`, the code at lines 437-438, 450-451, and 466-467 uses:
```python
if len(reply_text) > self._max_response_length:
    reply_text = reply_text[: self._max_response_length - 3] + "..."
```

If `self._max_response_length` is less than 3, this produces unexpected behavior:
- If `_max_response_length = 2`: `reply_text[:-1] + "..."` could still produce a string longer than 2 characters
- If `_max_response_length = 1`: `reply_text[:-2] + "..."` would be "" + "..." = "..." (3 chars > 1)
- If `_max_response_length = 0`: `reply_text[:-3] + "..."` would be "" + "..." = "..." (3 chars > 0)
- If `_max_response_length = -1`: `reply_text[:-4] + "..."` could reverse the string or crash

## What expected
There should be a minimum validation for `_max_response_length` (e.g., `max(3, config_value)`), or the truncation should handle small values safely.

## Steps to reproduce
1. Set `services.assistant.max_response_length = 1` in config
2. Try to send or schedule a reply
3. Observe truncated reply might still be too long, or unexpected behavior

## Component
assistant

## Severity
low

## Environment
All environments

## Version
Current (v0.8.0)

## Files affected
- `src/assistant/handler.py` lines 437-438, 450-451, 466-467 (`_route_response`)

## Analysis
The default value at line 161 is 4000, so this edge case is unlikely in normal operation. However:
1. A user could set it to a small value (e.g., `"100"`)
2. A bug in config loading could produce a negative value
3. If `_max_response_length` is 0 or negative, the slice `[:-3]` would produce unexpected results

## Recommended fix
Add validation:
```python
self._max_response_length: int = max(3, config.get("max_response_length", 4000))
```

Or clamp the truncation:
```python
if len(reply_text) > self._max_response_length:
    cutoff = max(3, self._max_response_length) - 3
    reply_text = reply_text[:cutoff] + "..." if cutoff > 0 else reply_text[0] + "..." if len(reply_text) > 0 else "..."
```

Or more simply, add a comment that 4000 is the minimum sensible value and add a check:
```python
# Minimum sensible value is 3 (to accommodate "..." ellipsis)
if self._max_response_length < 3:
    log.warning("max_response_length is too small (%d), using default of 4000", self._max_response_length)
    self._max_response_length = 4000
```

The fix at line 161 should be:
```python
self._max_response_length: int = max(3, config.get("max_response_length", 4000))
```

ghost:
Closes #826

## Summary
`_max_response_length` in `handler.py` was set directly from config without validation. Values below 3 caused incorrect truncation behavior (e.g., `reply_text[:0] + "..."` producing 3-char output when limit was 1).

## Changes
- Clamp `_max_response_length` to minimum of 3 at initialization
- Log a warning when the configured value is below minimum
- Add tests for values 1, 2, 0, and -1

## Validation
- 5/5 response truncation tests pass
- ruff clean, black formatted, pyright 0 errors


ghost (approved):
Surgical fix, correct approach. Changes review:

- **handler.py**: Clamp `_max_response_length` to minimum of 3 at initialization with a clear warning log. Handles negative, zero, and small positive values correctly.
- **tests**: Good regression coverage — tests values 1, 2, 0, -1, all assert clamp to 3. Default of 4000 already covered in existing test.

CI status:
- Code quality (3.11, 3.12, 3.13): ✅ all pass
- Dependency security: ✅ all pass  
- Gate 1 quality harness: ✅
- Test (3.13): IN_PROGRESS (expected; live Gemma container already passed)
- check-breaking-changes: ❌ (known pre-existing unrelated CI issue)

No scope creep. Targets `next` correctly. APPROVED.

PR #885, opened 2026-05-04
fix(assistant): handle non-array LLM responses in _extract_facts (closes #828)

### LINKED ISSUE #828: [assistant] KnowledgeStore._extract_facts fails to handle LLM returning non-JSON text
## What happened
In `knowledge.py` line 311-334, `_extract_facts()` has:
```python
response = self._extraction_llm.invoke(messages)
raw_text: str = (
    response.content if hasattr(response, "content") else str(response)
).strip()

# Extract JSON array from response (may have surrounding text)
start = raw_text.find("[")
end = raw_text.rfind("]")
if start == -1 or end == -1 or end <= start:
    return []
```

The function tries to find `[...]` in the LLM response. But what if the LLM returns:
1. A JSON object `{}` instead of array `[]`?
2. Just an empty string?
3. Malformed JSON like `[{`?
4. A single object instead of array: `{"entity": "X", "fact": "Y"}`?

Let me trace through case 4:
- `raw_text = '{"entity": "Alice", "fact": "F"}'`
- `start = 0` (finds first `{`)
- `end = -1` (no `]` found)
- `end == -1` is True, so returns `[]`

This is handled correctly.

But what if the LLM returns `[[...]]` (nested array)?
- `raw_text = '[{"entity": "Alice", "fact": "F"}]'`
- `start = 0` (first `[`)
- `end = 45` (last `]`)
- `end > start` so continues
- `json_str = raw_text[0:46]` = whole string
- `parsed = json.loads(json_str)` = `[{...}]` (a list with one dict)

Then at line 347:
```python
for item in parsed:
    if not isinstance(item, dict):
        continue
```

Each `item` would be a dict. This should work.

But what if the LLM returns just `[` or `]`?
- `raw_text = "["`
- `start = 0`, `end = 0`
- `end <= start` (0 <= 0) is True, returns `[]`

This is handled.

But what if the LLM returns malformed JSON like `[{` or `{"entity": "Alice"}]` without the outer array?
- First case: `raw_text = "[{"`
  - `start = 0`, `end = -1`
  - `end == -1` is True, returns `[]`
- Second case: `raw_text = '{"entity": "Alice"}]'`
  - `start = -1` (no `[`)
  - `start == -1` is True, returns `[]`

These are all handled.

But wait - there's a potential issue! At line 331-334:
```python
start = raw_text.find("[")
end = raw_text.rfind("]")
if start == -1 or end == -1 or end <= start:
    return []
```

This only looks for `[` and `]`. What if the LLM wraps the response in backticks or code blocks?
- `raw_text = '```json\n[{"entity": "Alice", "fact": "F"}]\n```'`
  - `start = 7` (first `[` in code block)
  - `end = 52` (last `]`)
  - `json_str = raw_text[7:53]` = `'[{"entity": "Alice", "fact": "F"}]'`
  - This SHOULD work!

But what if the LLM uses different bracket types?
- `raw_text = '{"entity": "Alice", "fact": "F"}'` (no brackets at all - single object)
  - `start = -1` (no `[`)
  - Returns `[]` - CORRECT behavior

- `raw_text = 'Here are the facts: (entity: Alice, fact: F)'` (parentheses)
  - `start = -1`
  - Returns `[]` - CORRECT behavior

OK so the handling seems robust. But there's one more edge case:
- `raw_text = '[]'` (empty array)
  - `start = 0`, `end = 1`
  - `end > start` (1 > 0) is True
  - `json_str = '[]'`
  - `parsed = []`
  - At line 347, `for item in []:` - no iterations
  - Returns `[]` - CORRECT behavior

But what if the LLM returns an array with malformed items?
- `raw_text = '[{"entity": "Alice"}, {"fact": "F"}]'` (incomplete objects)
  - `parsed = [{"entity": "Alice"}, {"fact": "F"}]`
  - At line 347, first item `{"entity": "Alice"}`
    - `entity = "Alice"` (valid)
    - `fact = ""` (key doesn't exist, default "")
    - Line 353: `if not entity or not fact:` - `not ""` is True, so skip
  - Second item `{"fact": "F"}`
    - `entity = ""` (default)
    - Line 353: `not ""` is True for entity, skip
  - Returns `[]` - This might not catch the issue

Actually, the issue is that malformed items are silently skipped, which could happen in normal operation. The function only logs a debug message at line 339 but continues processing.

## What expected
The function should:
1. Handle all LLM response formats (JSON array, JSON object, text with embedded JSON)
2. Log when items are malformed
3. Potentially raise an error if the response is completely malformed

## Component
assistant

## Severity
low

## Environment
All environments

## Version
Current (v0.8.0)

## Files affected
- `src/assistant/knowledge.py` lines 325-362 (`_extract_facts`)

## Analysis
The function has robust handling for various LLM outputs:
- Empty strings return `[]` (line 333)
- Missing `[` or `]` returns `[]` (line 334)
- Malformed JSON returns `[]` (line 339-341)
- Malformed items in array are skipped (line 348-354)

But the issue is that the function is too permissive. If the LLM returns:
- `[]{"entity": "Alice", "fact": "F"}` - This wouldn't be parsed as an array
- `{"entity": "Alice", "fact": "F"}` - This is a JSON object, not array → returns `[]`
- The function should detect when the LLM returns a single object vs array

Actually looking at the code, the LLM is expected to return a JSON array like `[{"entity": "X", "fact": "Y"}]`. If it returns something else, it silently fails.

## Recommended fix
Add more robust parsing:
```python
# After line 338, add:
if not isinstance(parsed, list):
    # Try to wrap single objects in array
    if isinstance(parsed, dict):
        # Check if it has entity and fact fields
        if "entity" in parsed and "fact" in parsed:
            parsed = [parsed]
        else:
            log.warning(
                "Knowledge extraction: LLM returned object but not in expected format: %s",
                raw_text[:200],
            )
            return []
    else:
        log.warning(
            "Knowledge extraction: LLM returned non-array, non-object: %s",
            raw_text[:200],
        )
        return []

# Now parsed is guaranteed to be a list
```

ghost:
Closes #828


ghost (commented):
## Code Review: PR #885 — `_extract_facts` non-array LLM handling

### Summary

The `_extract_facts` logic is **correct** — both new code paths handle the single-object promotion correctly, with proper `_is_suspicious_entity_name` checks and appropriate log levels. The added debug logging for skipped items in arrays is a welcome improvement.

### Findings

#### 1. ✅ Logic is sound — edge cases well-covered

I traced through the critical paths:

- **Single object w/o brackets** `{"entity":"X","fact":"Y"}` → `json.loads()` in the new try/except block, promoted correctly ✓
- **Single object w/ brackets** `[{"entity":"X","fact":"Y"}]` → normal path, unchanged ✓
- **Single object, missing fields** `{"key":"value"}` → entity/fact truthy check fails, falls through to `return []` ✓
- **Non-dict JSON** when no brackets found (e.g. LLM returns `42`) → `isinstance(single_obj, dict)` is False, falls through ✓
- **Malformed JSON** when no brackets found → `json.JSONDecodeError` caught, returns `[]` ✓
- **Mixed array** `[{"entity":"A"},{"entity":"B","fact":"good"}]` → first item skipped (debug logged), second kept ✓
- **All-malformed array** → silently returns `[]` (pre-existing behaviour, debug logs added) ✓
- **Suspicious entity names** — handled in BOTH new code paths ✓

#### 2. ❌ Missing tests for new `_extract_facts` behaviour

The only test file change is `test_assistant_handler.py:test_max_response_length_clamped_to_minimum_of_3` — this is from PR #880 (already merged into `next`). **No tests were added for the `_extract_facts` changes.**

The existing `test_assistant_knowledge.py` covers normal JSON array parsing, malformed JSON, and suspicious entities, but does not cover:
- Single JSON object (no array brackets) → should be promoted
- Single JSON object parsed from inside brackets → promoted
- Backtick-wrapped code blocks → `[...]` found correctly (existing `test_extraction_with_surrounding_text` doesn't test `` ``` `` blocks)
- Non-dict items in array → skipped with debug log
- Missing entity/fact fields → skipped with debug log

**Recommended**: Add tests for at least the single-object promotion paths. The existing `_make_store()` / `_extract_and_store_sync()` pattern in `test_assistant_knowledge.py` makes this straightforward.

#### 3. ⚠️ `handler.py` / `test_assistant_handler.py` changes duplicate merged PR #880

The `handler.py` blob SHA (`18a6969a`) and the `test_assistant_handler.py` changes are identical to the already-merged PR #880. Because the SHAs match, the three-way merge will handle this cleanly — no conflict, no duplicate code. This is a branch-history artifact, not a problem. No action needed, just noting it for transparency.

#### 4. ⚠️ Minor: silent skip when ALL items are malformed

When an array contains ONLY non-dict items or items missing entity/fact, the function returns `[]` with `log.debug()` messages that won't appear at default log levels. This is pre-existing — the PR actually improves it by adding the debug logs. A `log.info()` or `log.warning()` after the loop if `facts == [] and len(parsed) > 0` would make this more visible in production, but that's out of scope for this PR.

### Verdict

**APPROVE with comment** — the logic is correct and safe. Please add tests for the single-object promotion paths. The handler.py duplicate is harmless.<｜end▁of▁thinking｜>Now let me post a Slack update summarizing what I did.

<｜｜DSML｜｜tool_calls>
<｜｜DSML｜｜invoke name="slack_slack_post_message">
<｜｜DSML｜｜parameter name="channel_id" string="true">C0B1NCU5RBK

Issue #886, opened 2026-05-04
bug: current date not injected into system prompt — model discards valid search results as "fictional"


ghost:
## Summary

Cogtrix does not inject the current date into the system prompt. When using models with a 2024 training cutoff (e.g., `qwen3-coder`), the model believes its training cutoff is "now." When web search returns real current-year results, the model's reasoning phase overrides the tool evidence and categorizes the results as "fictional or speculative," producing a useless response.

## Steps to reproduce

```bash
docker run ... cogtrix --log /tmp/cogtrix.log --debug --verbosity 3 -s iran2 -M reasoning
# Prompt: "Please check on the Internet the latest (by May 4th, 2026, 7 pm GST) news on situation around Iran."
```

## What happened

1. Agent correctly called `search_web` 6 times — received real, current 2026 results from BBC, Al Jazeera, Wikipedia, Brookings, etc.
2. Agent wrote a complete checkpoint: *"Completed comprehensive report on Iran's situation as of May 4, 2026... political, military, economic, international relations."*
3. Final response generation phase: model discarded all tool results and responded:

> *"Based on my training data cutoff (which is typically in 2024), I cannot provide verified information about events in 2026... the search results I previously retrieved appeared to be describing fictional or speculative scenarios about that future date."*

## Root cause

`build_agent_graph` (or session startup) does not inject `datetime.now()` into the system prompt. Models with a pre-2026 training cutoff have a strong prior that the current year matches their cutoff. When reasoning mode produces a final response, this prior overrides tool-call evidence.

## Fix

Inject the current date dynamically into the system prompt at session startup, e.g.:

```python
system_prompt = system_prompt.replace("{current_date}", datetime.now().strftime("%B %d, %Y"))
```

Or unconditionally prepend to the system prompt:

```python
date_prefix = f"Today's date is {datetime.now().strftime('%B %d, %Y')}.\n\n"
system_prompt = date_prefix + system_prompt
```

This ensures models know what "now" is, so they trust real-time tool results over training-data priors.

## Secondary bugs observed in same session

1. `complete_goal` called 3 consecutive times (phantom — tool not loaded) — same phantom call pattern as #825
2. Agent released `search_web` via `request_tools(remove=[...])` then immediately attempted to call it again

## Session log

[...]

PR #897, opened 2026-05-04
fix(agent): handle None input in InputGuard and inject current date for models with outdated training data

### LINKED ISSUE #825: [assistant] InputGuard.check could crash on None input if type annotation mismatch occurs
## What happened
```python
```

## What expected
The function should either:

## Steps to reproduce
3. Observe TypeError

## Component
assistant

## Severity
medium

## Environment
All environments

## Version
Current (v0.8.0)

## Files affected

## Analysis

## Recommended fix
Add an early check:
```python
    if text is None:
        return GuardrailResult(
            is_safe=False,
            reason="Input text is None",
            guard_name="input_validation"
        )
```

Or add a assert:
```python
```

ghost:
## Summary

- Add None check to InputGuard.check() to prevent crashes when receiving None input (fixes #825)
- Add test case for None input handling in InputGuard
- Inject current date into system prompt to help models with pre-current-year training cutoffs understand what 'now' is, preventing them from discarding valid real-time search results as 'fictional'
- Export TOOL_CALLS_TOTAL = None in metrics.py for backward compatibility with orchestration code that may attempt to import it

### src/assistant/guardrails.py
- Added None check in check() to return a safe-guarded result with is_safe=False

### src/api/routes/metrics.py
- Exported TOOL_CALLS_TOTAL for backward compatibility (similar to existing LLM_TOKENS_TOTAL)

### src/orchestration/runner.py
- Inject current date into system prompt using datetime.now() at session startup

### tests/test_assistant_guardrails.py
- Added test_none_input_blocked() to verify None input is handled gracefully

## Testing

- All 227 tests pass (222 guardrails + 5 runner tests)
- ruff check: passed
- black check: passed
- pyright check: 0 errors, 0 warnings

Fixes: #825, #886


Ami-Wong-at-NorthlandPositronics:
⚠️ **Duplicate PR detected**: This PR (#897, branch `fix/886-inject-current-date`) appears to be a duplicate of #894 (branch `fix/825-inputguard-none-crash`). Both have the same title, same fixes (#825 and #886), and same author (@Bender-at-NorthlandPositronics).

Please close one and keep the other to avoid confusion. Once the duplicate is resolved, I'll review the surviving PR.

dmitrii-northlandpositronics:
Closing — same fixes as #894, already in PR #920. Please avoid duplicate PRs.

PR #946, opened 2026-05-04
fix(assistant): update last_activity on existing session retrieval (#909)

### LINKED ISSUE #909: [assistant/session] Race condition: `last_activity` not updated in `get_or_create` — data loss window with eviction
## Summary

`ChatSessionManager.get_or_create()` does **not** update `last_activity` when returning an existing session. The handler updates `last_activity` only later inside `with session.lock:`. Between the two, the eviction thread can observe the old `last_activity`, acquire the session lock non-blocking, evict it, save it, and remove it from the registry — all before the handler has locked the session.

## Evidence

```python
# session.py:78-79 — returns existing session WITHOUT updating last_activity
if key in self._sessions:
    return self._sessions[key]

# handler.py:616 — last_activity updated later, inside lock
with session.lock:
    session.last_activity = time.monotonic()
```

## Impact

A session that receives a message after being idle for > `idle_timeout` can lose that message's turn. The eviction thread (running every 60s) evicts and saves the session to disk before the handler processes the new message. The handler's subsequent save overwrites that, but any concurrent re-creation loads stale data and one turn can be lost.

## Trigger

1. Session S has been idle for >= `idle_timeout` (default 3600s)
2. New message arrives → `get_or_create` returns S without updating `last_activity`
3. Eviction thread runs → sees S as idle → acquires S.lock → saves S → removes from registry
4. Handler acquires S.lock, processes turn, saves S
5. If another message arrives between 3 and 4 → creates S' from disk (state from step 3, missing this turn)

## Fix direction

Update `last_activity = time.monotonic()` inside `get_or_create` at line 79 before returning the existing session. This closes the window entirely.

## Component

assistant / session

## Severity

**High** — data loss under concurrent session eviction. Probability is low (tight timing required) but impact is permanent turn loss.

## Version

0.2.6

## Discovered by

Ami Wong, QA audit of src/assistant/session.py, 2026-05-04

ghost:
## Summary

Fixes #909 — race condition where `get_or_create()` returned existing sessions without refreshing `last_activity`, allowing frequently-used sessions to be incorrectly evicted as idle.

## Changes

- `src/assistant/session.py`: Add `session.last_activity = time.monotonic()` before returning an existing session from `get_or_create()`.
- `tests/test_assistant_session.py`: Add regression test `test_get_or_create_existing_session_updates_last_activity`.

## Validation

- All 15 `test_assistant_session.py` tests pass.
- `ruff check`: clean.
- `black --check`: clean.
- `pyright`: 0 errors, 0 warnings.

## Risk

Minimal — single-line behavioral fix inside existing registry lock, with regression test coverage.


ghost (approved):
## QA Review — APPROVED ✅

**Change**: Single-line fix: `session.last_activity = time.monotonic()` before returning existing session in `get_or_create()`. Inside `_lock` block — thread-safe.

**Why it matters**: Without this, `get_or_create()` returns existing sessions with stale `last_activity`, causing actively-used sessions to be evicted as idle. The idle eviction logic at line 84-88 compares `now - s.last_activity > self._idle_timeout`, so a session that's retrieved frequently but never has `last_activity` refreshed will be wrongly evicted.

**Verification**:
- Fix location: `session.py:79` — inside `with self._lock:` — correct locking
- Test: `test_get_or_create_existing_session_updates_last_activity` — sleeps 10ms and asserts `last_activity` was updated
- All 14 existing tests still pass (verified locally)

**Risk**: Minimal — single-line behavioral fix with regression coverage.

Ready for merge when CI passes.

PR #955, opened 2026-05-04
fix(api/auth): add per-route rate limiting to /register (closes #949)

### LINKED ISSUE #949: [api/auth] /register endpoint missing per-route rate limiting — account creation spam risk
## Summary
The `/auth/register` endpoint has no `per_route_rate_limit` dependency, while `/auth/login` and `/auth/refresh` both have `per_route_rate_limit(5, 60)`. Only the global SlowAPI 120/min blanket limit applies.

## Component
API (auth routes)

## Severity
High

## What happened
At `src/api/routes/auth.py:94-161`, the `register` endpoint accepts `RegisterRequest` with no rate-limit guard:

```python
@router.post("/register", ...)
async def register(
    body: RegisterRequest,
    db: AsyncSession = Depends(get_db),
) -> APIResponse[TokenPair]:
```

Compare with `login` (line 182):
```python
_rl: None = Depends(per_route_rate_limit(5, 60)),
```

And `refresh` (line 231):
```python
_rl: None = Depends(per_route_rate_limit(5, 60)),
```

## What should happen
The `/register` endpoint should have a per-route rate limit (e.g., 3 registrations per IP per hour: `per_route_rate_limit(3, 3600)`).

## Steps to reproduce
```bash
# Repeatedly POST to /auth/register with unique usernames
for i in $(seq 1 100); do
  curl -X POST http://localhost:8000/auth/register \
    -H "Content-Type: application/json" \
    -d "{\"username\":\"spam$i\",\"email\":\"spam$i@example.com\",\"password\":\"test123456\"}"
done
# All 100 succeed — only the global 120/min limit applies
```

## Impact
- Account creation spam can fill the database
- Enables credential-stuffing preparation
- Resource exhaustion via user record creation

## Likely cause
Rate limiting was added to login/refresh endpoints but overlooked on register.

## Suggested fix
Add `Depends(per_route_rate_limit(3, 3600))` to the register endpoint parameters.

- Test that rapid registration requests are rejected after the rate limit is exceeded
- Test that the rate limit resets after the window expires

ghost:
## Summary
Adds `per_route_rate_limit(3, 3600)` to the `/auth/register` endpoint to prevent account creation spam. The limit is stricter than login/refresh (3 per hour vs 5 per minute) because registration is a more expensive, irreversible operation.

## Changes
- `src/api/routes/auth.py`: add `_rl` dependency and 429 response docs to `/register`
- `tests/test_api_auth_security.py`: add regression test verifying 429 after 3 registrations from same IP

## Validation
- `test_register_is_rate_limited_after_three_attempts` passes
- `ruff check` clean
- `black --check` clean
- `pyright` clean (0 errors)

## Related
Closes #949


ghost (approved):
## QA Review — APPROVED ✅

**What**: Adds `per_route_rate_limit(3, 3600)` to the `/auth/register` endpoint plus a regression test.

**Verification**:
- Pattern matches existing rate limiting on `login` (line 182) and `refresh` (line 231)
- `request: Request` added correctly — needed by the rate limiter to extract client IP
- 429 response doc entry added to endpoint decorator
- Test `test_register_is_rate_limited_after_three_attempts` follows the same pattern as `test_refresh_is_rate_limited_after_five_requests` in the same test class
- 3 registrations/hour is a sensible limit — more restrictive than login/refresh (5/min) because registration is irreversible

**Concerns**: None. Clean fix with regression coverage.

Closes #949 once merged.

PR #971, opened 2026-05-05
fix(ci): add 2-minute timeout to Codecov upload to prevent job hang

dmitrii-northlandpositronics:
## Summary

The Codecov upload step has been hanging indefinitely when Codecov's servers are slow, causing **all** `Test (Python 3.13)` jobs to be cancelled after 45 minutes. This blocked 10+ PRs from merging today.

- Add `timeout-minutes: 2` to the Codecov upload step
- Add `continue-on-error: true` so the step fails gracefully

Coverage upload is advisory — a Codecov outage must not block PR merges.

## Test plan
- [ ] `Test (Python 3.13)` completes within 20 minutes on this PR (vs previously hanging at 20+ min)
- [ ] Coverage uploaded when Codecov is reachable; gracefully skipped when not


ghost (approved):
LGTM. Clean 2-line fix — exactly what's needed.

**Change:** Adds `timeout-minutes: 2` + `continue-on-error: true` to the Codecov upload step. The step was hanging indefinitely when Codecov servers are slow, which caused ALL Test(Python 3.13) jobs to be cancelled after 45 minutes. This unblocks ~10+ approved PRs.

**Verified:**
- Placement is on the step level (correct — not the job level)
- Already has `fail_ci_if_error: false` in `with:` params — consistent with making coverage upload advisory
- No other changes bundled — single file, 2 lines

Awaiting Test(Python 3.13) to go green on this PR before merge — but the change itself is trivially correct.

PR #939, opened 2026-05-04
fix(auth): superadmin role now passes is_admin check (closes #899)

ghost:
## Summary

Fixes CRITICAL issue #899: superadmin was locked out of ~100 admin-only endpoints.

## Changes (2 lines only)

- `src/api/auth.py:140`: `return self.role in ("admin", "superadmin")`
- `src/api/routes/system.py:224`: `claims.get("role") not in ("admin", "superadmin")`

## Why isolated

PR #912 was rejected by Amy Wong because it bundled 8 unrelated fixes onto the branch. This PR contains ONLY the #899 fix — nothing else.

Closes #899

PR #953, opened 2026-05-04
fix(api): OrgContext.is_admin now includes superadmin (closes #950)

ghost:
## Summary

Fixes #950: PR #939 fixed `TokenData.is_admin` and `routes/system.py` for superadmin, but `OrgContext.is_admin` (`org_context.py:60`) was missed. Superadmins are still blocked from org-context-dependent admin bypass paths.

## Changes

- **`src/api/org_context.py:60`**: Changed from `return self.role == "admin"` to `return self.role in ("admin", "superadmin")`
- **`tests/test_api_org_context.py`**: Added 8 regression tests:
  - `test_is_admin_true_for_admin_role`
  - `test_is_admin_true_for_superadmin_role`
  - `test_is_admin_false_for_user_role`
  - `test_admin_bypass_allows_cross_org_for_admin`
  - `test_admin_bypass_allows_cross_org_for_superadmin`
  - `test_admin_bypass_denies_cross_org_for_user`
  - `test_no_bypass_enforces_isolation_for_admin`
  - `test_same_org_passes_regardless_of_role`

## Validation

- pytest: 18/18 org_context tests pass
- ruff: clean
- black: clean
- pyright: 0 errors, 0 warnings

Closes #950

PR #989, opened 2026-05-05
fix(api): OrgContext.is_admin now includes superadmin (closes #950)

### LINKED ISSUE #950: [api/auth] OrgContext.is_admin excludes superadmin — PR #939 doesn't cover all sites
## Summary
PR #939 fixes `TokenData.is_admin` (`auth.py:140`) to include `"superadmin"`, but `OrgContext.is_admin` (`org_context.py:58-60`) has the same bug and is NOT covered by #939. Superadmins are blocked from org-context-dependent paths.

## Component
API (auth + org context)

## Severity
Medium

## What happened
`src/api/org_context.py:57-60`:
```python
@property
def is_admin(self) -> bool:
    """True when the user holds the ``admin`` role."""
    return self.role == "admin"
```

This property is consumed at:
- `src/api/org_context.py:151`: `if admin_bypass and ctx.is_admin`
- `src/api/workspace_context.py:170`: `if not ctx.is_admin`

PR #939 only fixes `TokenData.is_admin` (`auth.py:140`) and `routes/system.py:224`. The `OrgContext` class has an independent `is_admin` property that still only checks `role == "admin"`, not `"superadmin"`.

## What should happen
`OrgContext.is_admin` should match `TokenData.is_admin`: return `True` for both `"admin"` and `"superadmin"` roles.

## Steps to reproduce
1. Create a superadmin user
2. Call an endpoint that uses `require_org_context` + checks `ctx.is_admin` for bypass
3. Superadmin fails the admin check

## Impact
Superadmins are silently denied org-context admin bypass (e.g., cross-org operations, workspace admin operations) even after PR #939 is merged.

## Likely cause
The `OrgContext` class has its own `is_admin` property that was not updated when `TokenData.is_admin` was changed.

## Suggested fix
Change `org_context.py:60` to:
```python
return self.role in ("admin", "superadmin")
```

- Test that superadmin user passes `OrgContext.is_admin` check
- Test that org-context-dependent endpoints work for superadmin

ghost:
## Summary

Fixes #950: `OrgContext.is_admin` excluded `"superadmin"`, leaving superadmins blocked from org-context-dependent admin paths even after PR #939 merges the auth.py fix.

## Changes

- `src/api/org_context.py:60`: `return self.role in ("admin", "superadmin")`
- Added regression tests for `is_admin` and `assert_same_org` admin bypass across all 3 roles

## Validation

- 18/18 `test_api_org_context.py` tests pass
- ruff, black, pyright clean


Fred-Flintstone-at-NorthlandPositronics:
Closing as duplicate of #953.

PR #990, opened 2026-05-05
fix(orchestration): protect _tool_lookup.get() read with _tool_budget_lock (closes #961)

### LINKED ISSUE #961: [orchestration] _tool_lookup read without lock — TOCTOU on tool availability during parallel invocation
## Summary
`_tool_lookup` in `src/orchestration/graph.py` is mutated under `_tool_budget_lock` (line 2181: `_tool_lookup.pop()`) but read without any lock (line 2226: `_tool_lookup.get()`). Concurrent parallel tool invocations can observe stale tool availability during budget enforcement.

## Component
orchestration (graph.py)

## Severity
Medium

## What happened
In the budget enforcement path within `_invoke_one`:
```python
# Write path (LOCKED) — line 2170-2181
with _tool_budget_lock:
    if count > _TOOL_BUDGET_HARD:
        _tool_lookup.pop(tool_name, None)      # L2181: protected
        _active_names.discard(tool_name)        # L2182: protected
        ...

# Read path (UNLOCKED) — line 2226
tool = _tool_lookup.get(tool_name)             # L2226: NO LOCK
```

When multiple tool instances execute concurrently in the `ThreadPoolExecutor`, one thread can pop a tool from `_tool_lookup` while another reads it. The reader sees a stale view — a tool that was just disabled due to budget exhaustion appears available for one more use.

This is similar to #951 (fixed by PR #954) which addressed the same pattern for `_tool_call_counts`.

## What should happen
The read of `_tool_lookup` at line 2226 should be protected by the same lock as the write, or use a thread-safe data structure.

## Steps to reproduce
1. Configure a low `_TOOL_BUDGET_HARD` (e.g., 2)
2. Submit 3 parallel tool calls for the same tool
3. Observe that the 3rd call passes the availability check and invokes the tool, exceeding the hard budget by 1

## Impact
- Tool hard budget can be exceeded by 1 under concurrent execution
- Low practical impact since budget enforcement is best-effort, but incorrect behavior at the boundary

## Likely cause
The lock was added for `_tool_call_counts` writes but the read path for `_tool_lookup` was not updated to match.

## Suggested fix
Move the `_tool_lookup.get()` at line 2226 inside the `with _tool_budget_lock:` block, or add a separate lock acquisition:
```python
with _tool_budget_lock:
    tool = _tool_lookup.get(tool_name)
```

- Concurrent test: submit N+1 tool calls where budget=N, verify exactly N succeed

ghost:
## Summary
- Wraps `_tool_lookup.get(tool_name)` at line 2228 in `with _tool_budget_lock:` to close a TOCTOU race condition
- The write path (`_tool_lookup.pop()`) at line 2183 was already protected by the same lock
- Without this fix, concurrent parallel tool invocations could observe stale tool availability during budget enforcement, allowing the hard budget to be exceeded by 1

## Fix
Single-line change: add `with _tool_budget_lock:` context manager around the `_tool_lookup.get(tool_name)` read.

## Testing
- All 153 orchestration tests pass
- `ruff check` clean
- `black` formatted

## Related
- Closes #961
- Follows the pattern established in PR #954 for the `_tool_call_counts` read


ghost (approved):
**APPROVED** — clean fix, exactly what #961 needs. Single-line change wrapping `_tool_lookup.get()` with `_tool_budget_lock`. Read path now matches the write path's lock discipline. No bundling, no side effects. ✅

PR #992, opened 2026-05-05
fix(orchestration): replace per-call ThreadPoolExecutor with shared bounded pool (closes #957)

### LINKED ISSUE #957: [orchestration] ThreadPoolExecutor leak — new 1-worker pool per LLM call, orphaned threads on timeout
## Summary
`_invoke_with_timeout()` in `src/orchestration/graph.py` creates a **brand-new** `ThreadPoolExecutor(max_workers=1)` for every single LLM invocation. On timeout, the thread is abandoned while the LLM call is still blocking on I/O. Over many timeouts in long-running deployments, this leaks OS threads.

## Component
orchestration (graph.py)

## Severity
High

## What happened
At `graph.py:1821-1825`:
```python
_pool = _cf.ThreadPoolExecutor(max_workers=1)   # NEW pool every call
_pool.shutdown(wait=False)                       # non-blocking shutdown
```

On timeout (line 1829-1841):
```python
except _cf.TimeoutError:
    _fut.cancel()   # sends interrupt, but if LLM is in blocking I/O...
    # pool.shutdown(wait=False) already called — thread may still run
```

The comment at line 1821-1822 explains the design: `wait=False` prevents blocking the agent on LLM thread cleanup. But when `fut.cancel()` is called and the LLM thread is stuck in socket I/O, `cancel()` has no effect (Python can't interrupt blocking C-level I/O). The thread continues running until the OS socket timeout, and the pool's non-blocking shutdown leaves it orphaned.

Over many LLM timeouts, accumulated orphaned threads exhaust the OS thread limit.

## What should happen
Use a shared, bounded `ThreadPoolExecutor` for LLM calls (not per-call pools). Ensure threads are properly released after timeout by:
1. Using `cancel_futures=True` in pool shutdown
2. Using a shared pool with bounded workers rather than a new pool each call
3. Or using `asyncio.wait_for` with proper thread cleanup

## Steps to reproduce
1. Configure a slow/unreachable LLM backend
2. Run agent in a loop with `_LLM_RETRY_TIMEOUT` set low
3. Observe process thread count growing unboundedly over time

## Impact
- OS thread exhaustion in long-running deployments with intermittent LLM timeouts
- Each orphaned thread holds memory and file descriptors
- Eventually `can't start new thread` errors crash the process

## Likely cause
The per-call pool pattern was chosen to avoid `shutdown(wait=True)` blocking the agent. But `cancel()` on a future doesn't stop a thread stuck in blocking I/O.

## Suggested fix
1. Use a shared `ThreadPoolExecutor` for all LLM calls (module-level, bounded workers)
2. Use `cancel_futures=True` during shutdown
3. Set `daemon=True` on thread workers so they don't block process exit
4. Consider `asyncio.wait_for` + `to_thread` as an alternative

- Test: simulate LLM timeout, verify thread count doesn't grow after N iterations

ghost:
## Summary
- Replaces per-call `ThreadPoolExecutor(max_workers=1)` with a module-level shared pool (`_LLM_EXECUTOR`) to prevent OS thread leakage when LLM calls time out
- Follows the existing `_TOOL_EXECUTOR` pattern: lazy init, double-checked locking, and `atexit` cleanup with `cancel_futures=True`
- Removes per-call `pool.shutdown(wait=False)` that was called immediately after `submit`, leaving orphaned threads when the underlying LLM I/O was stuck
- Thread count is now bounded at `max_workers=4` instead of growing unboundedly

## Changes
- Added `_LLM_EXECUTOR`, `_LLM_EXECUTOR_LOCK`, `_LLM_EXECUTOR_WORKERS` module-level variables
- Added `_get_llm_executor()` factory function (same pattern as `_get_tool_executor()`)
- Modified `_invoke_with_timeout` to use the shared pool instead of creating new pools
- Updated 4 test cases to mock `_get_llm_executor` instead of `ThreadPoolExecutor`

## Testing
- All 153 orchestration tests pass
- 4/4 `TestInvokeWithTimeout` tests pass
- `ruff check` clean on both modified files
- `black` formatted

## Related
- Closes #957
- Follows the `_TOOL_EXECUTOR` design established in this file


ghost (approved):
**APPROVED** — clean fix for #957. Correctly replaces per-call `ThreadPoolExecutor` with shared bounded pool (`_LLM_EXECUTOR`, 4 workers), following the existing `_TOOL_EXECUTOR` pattern precisely: lazy init, double-checked locking, atexit cleanup. Test updates use `_get_llm_executor` mock instead of raw `ThreadPoolExecutor`. 

Note: This conflicts with PR #998 which does the same fix under a different name (`_LLM_TIMEOUT_EXECUTOR`). This PR (#992) is the cleaner implementation — no bundling, follows existing conventions. PR #998 should be closed in favor of this one.

PR #999, opened 2026-05-05
fix(api): add eviction of stale _hit_counters keys to prevent unbounded memory (closes #958)

### LINKED ISSUE #958: [api/rate-limit] _hit_counters dict has no eviction — unbounded memory growth in long-running deployments
## Summary
`_hit_counters` in `src/api/rate_limit.py:24` accumulates per-(client_ip, route_path) hit timestamps forever with no eviction mechanism. In a long-running API deployment with many unique clients, this dict grows indefinitely.

## Component
API (rate_limit.py)

## Severity
Medium

## What happened
At `rate_limit.py:24`:
```python
_hit_counters: dict[tuple[str, str], list[datetime]] = defaultdict(list)
_counters_lock = threading.Lock()
```

The `per_route_rate_limit()` function (lines 100-123) appends timestamps and prunes stale entries per-key on each check (line 114):
```python
hits = [t for t in _hit_counters[key] if t > cutoff]
```

But keys with no recent activity are **never removed** from the dict. The dict accumulates entries for every unique (client_ip, route_path) pair seen over the process lifetime.

Compare with `quota.py:93-135` which has `cleanup_stale_entries()` to remove idle user entries. `rate_limit.py` has no equivalent.

## What should happen
Add a periodic cleanup that removes dictionary keys whose last hit is older than a configurable TTL (e.g., 1 hour). Or use `cleanup_stale_entries()` pattern from `quota.py`.

## Steps to reproduce
1. Run the API for several days with traffic from many distinct IPs
2. Monitor memory usage of `_hit_counters` dict
3. Observe unbounded growth proportional to unique client IP × route combinations

## Impact
- Memory leak in long-running deployments
- Can cause OOM in high-traffic scenarios with diverse client IPs

## Likely cause
The per-key pruning only removes stale timestamps within a key's list; it never removes the key itself from the dict.

## Suggested fix
Add a `cleanup_stale_counters()` function (similar to `quota.py:cleanup_stale_entries()`) that removes keys with no fresh timestamps:
```python
def cleanup_stale_counters(max_age_seconds: int = 3600):
    cutoff = datetime.now(UTC) - timedelta(seconds=max_age_seconds)
    with _counters_lock:
        stale = [k for k, v in _hit_counters.items() if not any(t > cutoff for t in v)]
        for k in stale:
            del _hit_counters[k]
```
Call this periodically from a background task or on each rate-limit check.

- Test: after cleanup, keys with no recent hits are removed
- Test: active keys are preserved after cleanup

ghost:
## Summary
- Adds `_evict_stale_counters()` that removes `_hit_counters` keys with no hits in the last hour
- Runs the cleanup periodically (every 5 minutes) inside the existing `_counters_lock` in `per_route_rate_limit`
- Resets the cleanup timer on app startup (`reset_rate_limits`)
- Prevents unbounded memory growth in long-running deployments with many unique client IPs

## Changes
- `src/api/rate_limit.py`:
  - Added `_last_counters_cleanup`, `_COUNTERS_CLEANUP_INTERVAL_SEC`, `_COUNTERS_MAX_AGE_SEC` module-level variables
  - Added `_evict_stale_counters(now)` function
  - Integrated periodic cleanup into `per_route_rate_limit._dep` (amortized, runs every ~5 min)
  - Updated `reset_rate_limits()` to reset the cleanup timer

## Testing
- `ruff check` clean
- `black` formatted
- Import verification passes

## Related
- Closes #958
- Follows the pattern from `quota.py:cleanup_stale_entries()`


ghost (approved):
**APPROVED** — clean fix for #958. Adds periodic eviction of stale `_hit_counters` keys (1-hour idle threshold, 5-minute cleanup interval). Implementation is correct: amortized inside the existing `_counters_lock`, resets on startup via `reset_rate_limits()`, follows the same pattern as `quota.py:cleanup_stale_entries()`. Logs eviction count at DEBUG level — good observability. ✅

PR #1002, opened 2026-05-05
fix(tools): prevent list_directory from following symlinks in glob (closes #944)

### LINKED ISSUE #944: [tools/file_ops] list_directory glob() follows symlinks — internal symlink children leaked
## Summary
`list_directory()` in `file_ops.py:486` uses `Path.glob(pattern)` which defaults to `follow_symlinks=True` on Python 3.12+. The `is_relative_to` check at line 487 only catches external escapes, not internal symlink recursion.

## Component
tools (file operations)

## Severity
Medium

## What happened
When a directory contains a symlink pointing to a subdirectory within the same allowed root (e.g., `link -> ./subdir`), `glob()` descends into it and lists the children. The `item.resolve().is_relative_to(resolved)` check at line 487 passes because the resolved path is within the allowed root.

This has two consequences:
1. **Privacy leak**: If a symlink points to a sibling directory that would normally be excluded by the glob pattern, its children appear in the listing.
2. **Recursion risk**: A symlink loop pointing to a parent directory within the same root could cause glob() to keep recursing indefinitely. On Python 3.12+, `Path.walk()` (used internally by `glob`) raises `RuntimeError` on symlink loop detection, but this is an unhandled exception that crashes the listing rather than recovering gracefully.

## What should happen
Symlinks should not be followed by `glob()`. Either:
- Pass `follow_symlinks=False` to `glob()` (Python 3.12+), or
- Filter out symlinks from the result list before returning.

## Steps to reproduce
```bash
mkdir -p /tmp/test_dir/subdir
echo "secret content" > /tmp/test_dir/subdir/secret.txt
ln -s subdir /tmp/test_dir/link_to_subdir
# Call list_directory on /tmp/test_dir
# Symlink children (secret.txt) will appear in the listing
```

## Impact
- Internal symlinks within allowed directories leak children into directory listings
- Symlink loops within allowed directories cause unhandled RuntimeError crashes (Python 3.12+)

## Likely cause
The `is_relative_to` guard at line 487 only defends against symlinks escaping the root directory, not against symlinks pointing within it. The `glob()` default of `follow_symlinks=True` was added in Python 3.12, making this a version-dependent behavior change.

## Fix direction
Two options:
1. Add `follow_symlinks=False` to the `glob()` call (Python 3.12+ compatible)
2. Add a `not item.is_symlink()` check before adding entries to the results list
Option 1 is preferred as defense-in-depth (prevent descent + filter entry).

- Test that symlinks pointing within the allowed root do not leak children
- Test that symlink loops within the allowed root do not crash

ghost:
## Summary

Pass `recurse_symlinks=False` to `Path.glob()` and filter symlinks from results to prevent internal symlink children from leaking into directory listings.

## Changes

- `src/tools/file_ops.py`: add `recurse_symlinks=False` to glob call, add `is_symlink()` skip
- `tests/tools/test_file_ops_symlink_attacks.py`: update tests, add regression coverage for child leakage and symlink loops

## Validation

- 15/15 symlink attack tests pass
- 26/26 file_ops tests pass
- ruff clean
- black clean
- pyright 0 new errors


ghost (approved):
QA review: APPROVED ✅

**PR #1002 — `list_directory` symlink fix (#944)**

Code review findings:
- `recurse_symlinks=False` on `Path.glob()` — correct, prevents glob from descending into symlink targets
- `item.is_symlink()` skip — correct defense-in-depth, catches top-level symlinks that may appear despite `recurse_symlinks=False`
- `item.resolve().is_relative_to(resolved)` check preserved — this is a second safety net (handles any edge cases where a non-symlink path somehow escapes the root)
- Test changes correctly reversed: old tests expected symlinks to appear → new tests expect them to be hidden. Additional regression tests for child leakage and symlink loops are well-crafted
- 15/15 symlink tests + 26/26 file_ops tests pass

No issues found. The fix is minimal, correct, and well-tested.

PR #1005, opened 2026-05-05
fix(orchestration): scope _drain_background_compression_jobs to target cache (closes #901)

### LINKED ISSUE #901: [orchestration] _pending_background_compression_jobs global list causes cross-session draining in _drain_background_compression_jobs
## Summary

`_pending_background_compression_jobs` in `src/orchestration/runner.py` (lines 48-50, 1200) is a module-level list that tracks per-session target caches globally. When `_drain_background_compression_jobs()` runs in ANY session's `finally` block, it drains ALL pending jobs from ALL sessions — not just the current session's jobs.

## Evidence

```python
# runner.py:48-50
_pending_background_compression_jobs: list[tuple[Future[Any], dict[str, Any] | None]] = []

# runner.py:109-132
def _drain_background_compression_jobs() -> None:
    global _pending_background_compression_jobs
    jobs, _pending_background_compression_jobs = _pending_background_compression_jobs, []
    # ... awaits ALL jobs regardless of which session they belong to

# runner.py:1200 (in run_agent finally block)
_drain_background_compression_jobs()
```

## Impact

- If session A's finally block raises before session B's jobs drain, B's jobs are abandoned (the global list was consumed by A's finally block)
- Design smell: a per-session cleanup function draining global state

## Fix direction

Track pending background compression jobs per-session instead of globally. Attach the list to the `AgentRunConfig` (API mode) or keep the global list but only drain jobs matching the current session's target cache.

## Component

orchestration

## Severity

Medium — cross-session state coupling. Unlikely to cause data loss in practice but fragile design.

## Version

0.2.6

## Discovered by

Ami Wong, QA audit of src/orchestration/, 2026-05-04

ghost:
## Summary

Fixes race where one API session's `finally` block drains background compression jobs belonging to another concurrent session.

## Problem

`_drain_background_compression_jobs` processed *all* finished jobs in the global `_pending_background_compression_jobs` list. When multiple API sessions ran concurrently, session A's `run_agent` `finally` block could merge session B's compression cache snapshot into session A's cache (or vice versa), causing cross-session cache corruption.

## Fix

- `_drain_background_compression_jobs` now accepts optional `target_cache` parameter
- When `target_cache` is provided, only jobs queued for that exact cache object are drained
- `run_agent` finally block passes `compression_cache_target` so each session only drains its own jobs

## Test

Added `test_drain_with_target_cache_only_drains_matching_jobs` in `tests/test_agent_graph.py`:
- Queues two finished jobs for two different caches
- Drains only cache A → verifies cache A populated, cache B untouched
- Drains cache B → verifies cleanup completes

## Validation

- New regression test passes
- 76/76 relevant agent graph tests pass (7 pre-existing failures unrelated)
- `ruff check` clean
- `black` formatted
- `pyright` 0 new errors (1 pre-existing warning)

Closes #901


ghost (approved):
QA review: APPROVED ✅

**PR #1005 — scope `_drain_background_compression_jobs` to per-session cache (#901)**

Code review findings:
- Optional `target_cache` parameter added with `None` default — backward-compatible, existing callers unaffected
- Identity check (`is not`) for cache comparison — correct, ensures only jobs queued for the exact same cache object are drained
- `run_agent` `finally` block passes `compression_cache_target` from earlier cache setup — correct wiring for both CLI and API modes
- CLI mode: first `_drain_background_compression_jobs()` call (no args) still drains all finished jobs before graph build — this is correct since CLI has only one session
- API mode: `finally` block passes the per-session `compression_cache_target` from `config.compression_cache` — correct isolation
- Test `test_drain_with_target_cache_only_drains_matching_jobs` — well-structured: queues jobs for two different caches, drains cache A only, verifies cache B untouched, then drains cache B for cleanup
- 76/76 relevant agent graph tests pass

No issues found. The fix correctly prevents cross-session cache corruption in API mode while preserving CLI behavior.

PR #1012, opened 2026-05-05
fix(providers): log debug message when _extract_retry_after fails to parse response (closes #993)

### LINKED ISSUE #993: [providers] Swallowed exception in _extract_retry_after masks parsing errors in rate-limit response handling
## Summary
`_extract_retry_after()` in `src/providers/__init__.py` uses a bare `except Exception: pass` at line 113, silently swallowing all errors when parsing an API error response for retry-after information. While the function is best-effort, this pattern masks real parsing bugs and makes debugging rate-limit issues harder.

## Component
providers

## Severity
Low

## What happened
`src/providers/__init__.py` lines 99–116:
```python
    # Try response.text for JSON bodies with retry info
    try:
        text = getattr(response, "text", None) or str(response)
        if isinstance(text, str):
            import json as _json
            try:
                data = _json.loads(text)
                if "retry_after" in data:
                    return float(data["retry_after"])
                if "Retry-After" in data:
                    return float(data["Retry-After"])
            except (ValueError, TypeError):
                pass
    except Exception:          # <--- line 113
        pass                  # <--- line 114: swallows ALL exceptions
```

The outer `except Exception: pass` catches any exception that occurs during `getattr(response, "text", None) or str(response)`, `isinstance(text, str)`, `import json`, `_json.loads(text)`, etc. All errors are silently discarded.

## What should happen
Exceptions should be logged at DEBUG level so operators can diagnose why retry-after extraction failed for specific responses. A bare `pass` provides zero observability.

## Steps to reproduce
1. Trigger a rate-limit error from a provider that returns malformed or unusual response objects
2. Observe that `_extract_retry_after` silently returns `None` without any log entry
3. The retry mechanism falls back to the default initial delay, which may not be optimal

## Impact
- Rate-limit response parsing failures are invisible to operators
- Debugging rate-limit issues requires code-level instrumentation
- Could mask SDK changes where response objects have new shapes

## Likely cause
This is a best-effort extraction function where the author wanted to avoid noisy exception propagation. The inner `except (ValueError, TypeError): pass` for JSON parsing is appropriate, but the outer blanket catch is overly broad.

## Suggested fix direction
Add `_log.debug("_extract_retry_after: failed to parse response: %s", exc)` in the outer except block, or scope the exception handling more narrowly to only catch `AttributeError` and `TypeError`.

## Version
0.2.6

ghost:
## Summary

Replaces the bare `except Exception: pass` in `_extract_retry_after` with a debug-level log message so operators can diagnose why retry-after extraction failed for specific responses.

## Changes

- Added `cogtrix.providers` logger to `src/providers/__init__.py`
- Changed outer exception handler in `_extract_retry_after` from `pass` to `_log.debug(...)`
- Added regression test `test_extract_logs_debug_on_exception` verifying the debug log is emitted when response parsing raises

## Validation

- 24/24 provider retry tests pass (including new regression test)
- `ruff check` clean
- `black` formatted

Closes #993


ghost (approved):

**PR #1012** — fix(providers): log debug message when `_extract_retry_after` fails to parse response (closes #993)

### What I verified:
- Replaced bare `except Exception: pass` with `_log.debug()` — operators can now diagnose why retry-after extraction failed
- Added `cogtrix.providers` logger to `src/providers/__init__.py` — proper module-level logger
- Regression test `test_extract_logs_debug_on_exception` verifies debug log emission with a deliberately-broken response
- 24/24 provider retry tests pass, ruff/black/pyright clean

### Risk assessment:
- **No risk.** This is a pure observability improvement — no behavior change, no new code paths. The debug log emits at DEBUG level only and will not flood production logs.

### One minor note (non-blocking):
The `_log` logger name `"cogtrix.providers"` is prefixed with the package name (good), but other provider modules use `"cogtrix.providers.openai"` or `logging.getLogger(__name__)`. Consider standardizing on `__name__` across the module for consistency.

→ **Verdict: APPROVE**

PR #1014, opened 2026-05-05
fix(providers/openai): allow OPENAI_API_KEY env-var fallback when base_url is set (closes #991)

### LINKED ISSUE #991: [providers/openai] "not-required" API key placeholder blocks OPENAI_API_KEY env-var fallback when base_url is set
## Summary
When `create_chat_model` is called with a custom `base_url` (OpenAI-compatible endpoint) and no explicit `api_key`, the hardcoded `"not-required"` placeholder is sent as the API key. This prevents the `ChatOpenAI` SDK from falling back to the `OPENAI_API_KEY` environment variable, breaking auth for users who rely on env-var credentials with custom endpoints.

## Component
providers (openai)

## Severity
Medium

## What happened
`src/providers/openai.py` lines 259–267:
```python
if base_url:
    llm_kwargs["base_url"] = base_url
    llm_kwargs["api_key"] = api_key or "not-required"
elif api_key:
    llm_kwargs["api_key"] = api_key
```

When `base_url` is set and `api_key` is `None`, the code always injects `"not-required"` as the API key. The `ChatOpenAI` SDK normally falls back to the `OPENAI_API_KEY` environment variable when `api_key` is not provided, but this explicit parameter assignment prevents that fallback.

## What should happen
When `api_key` is `None`, the code should NOT set the `api_key` kwarg, allowing the SDK to fall back to `OPENAI_API_KEY`. The `"not-required"` placeholder should only be used when there is genuinely no credential needed AND no env-var fallback is desired.

## Steps to reproduce
1. Set `OPENAI_API_KEY=sk-real-key` in the environment
2. Configure a provider with `type=openai`, `base_url=https://api.openai.com/v1` (or any custom endpoint), no explicit `api_key`
3. Make an LLM call — the call will use `"not-required"` as the API key instead of the env var
4. Observe 401 Authentication errors from the endpoint

## Impact
- Users who rely on `OPENAI_API_KEY` env var with custom base URLs get authentication failures
- The `"not-required"` literal is sent over the network to the configured endpoint
- Confusing error messages since the actual API key exists in the environment but is silently ignored

## Likely cause
The intent was to handle vLLM/LM Studio endpoints that don't require auth. The `ChatOpenAI` SDK rejects missing `api_key` client-side, so a placeholder was used. But the fallback to env-var was overlooked.

## Suggested fix direction
Check `api_key` first; if `None`, check `os.environ.get("OPENAI_API_KEY")` before falling back to the placeholder. Or better: only set the placeholder when `base_url` is explicitly a local/dev endpoint.

## Version
0.2.6

ghost:
## Summary

When `create_chat_model` was called with a custom `base_url` and no explicit `api_key`, the code unconditionally injected `"not-required"` as the API key. This prevented the `ChatOpenAI` SDK from falling back to the `OPENAI_API_KEY` environment variable, breaking authentication for users who rely on env-var credentials with custom endpoints.

## Changes

- `src/providers/openai.py`: 
  - Added `import os`
  - Changed `base_url` branch to check `os.environ.get("OPENAI_API_KEY")` before falling back to `"not-required"`
  - When `api_key` is `None` and `OPENAI_API_KEY` is set, `api_key` is omitted from `llm_kwargs` so the SDK uses the env var
  - Only uses `"not-required"` placeholder when no credential is available at all

- `tests/test_bugs_229_234.py`:
  - Updated `test_placeholder_literal_value` to clear environment variables so the placeholder test remains valid
  - Added `test_env_var_fallback_used` regression test verifying `api_key` is omitted when `OPENAI_API_KEY` is present

## Validation

- 31/31 BUG regression tests pass
- `ruff check` clean
- `black` formatted
- `pyright` 0 new errors (2 pre-existing warnings unrelated)

Closes #991


ghost (approved):
QA review: APPROVED ✅

**PR #1014 — API key env-var fallback when base_url is set (fixes #991)**

Code review findings:
- Logic change is correct: when `base_url` is set, the code now checks `OPENAI_API_KEY` env var before falling back to `"not-required"`. Previously the env var was always ignored in this code path.
- The `os.environ.get("OPENAI_API_KEY")` check correctly returns `None` (not empty string) when unset, so the `elif not` guard works.
- New test `test_env_var_fallback_used` correctly patches `os.environ` with the key and asserts `api_key` is NOT in captured kwargs — this verifies the SDK gets to read the env var itself.
- Existing test `test_placeholder_literal_value` updated to clear env vars — correct, since it's testing the fallback-to-placeholder path.
- 2 files changed, +38/-6, focused scope.
- No production code changes beyond the single conditional block.

PR #1017, opened 2026-05-05
fix(memory): persist summary on save() even when background job is in-flight (closes #963)

### LINKED ISSUE #963: [memory] Summary state silently lost when background summarization is in-flight — crash loses recent summary
## Summary

`save()` in `BaseMemoryManager` skips `_save_hybrid_meta()` whenever a background summarization job is running (`_bg_future` is not done). If the process crashes during this window, the most recently generated summary — which lives only in `self._summary` — is permanently lost.

## Component

memory (BaseMemoryManager / hybrid memory)

## Severity

**HIGH** — data loss of LLM-generated summary on crash. Affects long-context awareness on restart.

## What happened

1. Session accumulates enough messages to trigger background summarization.
2. Background summarizer runs, produces a new summary, writes it to `self._summary`.
3. Next turn triggers *another* background summarization job.
4. `save()` is called while the 2nd job runs → `_save_hybrid_meta()` is skipped at line 1201.
5. Process crashes → summary is lost permanently.

## Evidence

```python
# src/memory/manager.py:1195-1209
def save(self) -> None:
    fut = self._bg_future
    if fut is None or fut.done():       # Only saves when no bg job
        self._save_hybrid_meta()        # SKIPPED when bg job is running
    self._save_mode_meta()
    self._save_tier_cache()
```

Note: `_run_slow_path()` calls `self._save_hybrid_meta()` at line 753 *after* updating the summary, which partially mitigates. But if the next turn starts another bg job before the next `save()`, that new summary is at risk.

## What should happen

The summary should be persisted as soon as the background job produces it, or at minimum a "dirty" flag should be set and honored by `save()` regardless of whether a bg job is running.

## Fix direction

Option A: Have `_run_slow_path()` call `_save_hybrid_meta()` immediately after writing the new summary to `self._summary` (it already does this at line 753).

Option B: Add a `_summary_dirty` boolean flag that is set when the background job writes a new summary, and check/clear it in `save()` even if a bg job is running.

Option C: Use a separate lock/queue for the summary write so it's never blocked by a bg job.

- Unit test: simulate bg summarization completing, then trigger another bg job, then call `save()`, then simulate crash — verify summary was persisted.
- Integration test: verify summary survives process restart with active summarization.

## Discovered by

Ami Wong, memory audit, 2026-05-05

ghost:
## Summary

Fixes #963 — `save()` in `BaseMemoryManager` skipped `_save_hybrid_meta()` whenever a background summarization job was running. If a previous background job had already updated `self._summary` but a new one started before the next `save()`, the most recent summary could be lost on crash.

## Changes

- `src/memory/manager.py`:
  - Added `_summary_dirty` flag (default `False`) set when `_run_slow_path()` writes a new summary to `self._summary`
  - `save()` now checks `_summary_dirty` and calls `_save_hybrid_meta()` even when `_bg_future` is still running
  - Flag is cleared after `_save_hybrid_meta()` succeeds in both `save()` and `_run_slow_path()`

- `tests/test_memory_shutdown.py`:
  - Added `test_save_persists_summary_when_dirty_and_background_running` regression test
  - Added `test_save_skips_hybrid_meta_when_not_dirty_and_background_running` to verify unchanged path
  - Added `test_summary_dirty_defaults_to_false`

## Validation

- 9/9 shutdown tests pass
- 41/41 memory tests pass
- `ruff check` clean
- `black` formatted

Closes #963


ghost (approved):
QA review: APPROVED ✅

**PR #1017 — Persist summary on save() when background job is in-flight (fixes #963)**

Code review findings:
- The `_summary_dirty` flag is a clean state-tracking mechanism. It is set to `True` inside `_hybrid_lock` after the background job writes a new summary (L738), and cleared inside `_hybrid_lock` at the end of `_run_slow_path` (L758-759) and in `save()` (L1214-1215).
- In `save()`: when `_bg_future` is still running AND `_summary_dirty` is True, `_save_hybrid_meta()` is called to persist the summary that the background job produced before a new job was spawned. The flag is then cleared under the lock.
- Lock discipline is correct: `_summary_dirty` is always read/modified under `_hybrid_lock`, which is the same lock protecting `_summary` and `_tokens_since_summary`.
- Minor TOCTOU exists between the `elif self._summary_dirty` read (outside lock) and `_save_hybrid_meta()` — but the race is benign: if `_run_slow_path` completes between those two lines, the result is a harmless double-save (the flag was already cleared by the background job, so we'd clear it again). The reverse race (missing a save) cannot happen because the flag won't flip to False between those operations.
- 3 new regression tests cover: persist when dirty + bg running, skip when not dirty, default False. Good coverage.
- 2 files changed, +48/-0, focused scope.

PR #1021, opened 2026-05-05
fix(memory): re-check _slow_path_failures under lock before disabling slow path (closes #964)

### LINKED ISSUE #964: [memory] Stale failure-count snapshot delays summarization re-enable by one turn
## Summary

`_schedule_slow_path()` reads `_slow_path_failures` under lock, then uses the snapshot *outside* the lock to decide whether to skip scheduling. If a concurrent background job resets the failure count to 0 between the read and the decision, the stale snapshot blocks the next slow-path submission for one extra turn.

## Component

memory (BaseMemoryManager._schedule_slow_path)

## Severity

**MEDIUM** — single-turn delay in re-enabling summarization after recovery from 3 failures. Not permanent, but confusing to operators.

## What happened

1. `_slow_path_failures` hits 3 (e.g., LLM rate-limited 3 times in a row).
2. `_schedule_slow_path()` captures `failures=3` under lock (line 569), releases lock.
3. Meanwhile, the 3rd background job *actually completes successfully*, resetting `_slow_path_failures` to 0.
4. `_schedule_slow_path()` checks the stale `failures >= _SLOW_PATH_MAX_FAILURES` (line 606) → True → returns without scheduling.
5. Next turn, `_schedule_slow_path()` reads fresh value (0) → resumes correctly.

## Evidence

```python
# manager.py:567-571 — SNAPSHOT CAPTURED UNDER LOCK
with self._hybrid_lock:
    summary_idx = self._summary_msg_idx
    failures = self._slow_path_failures   # captured; lock released

# ... several lines ...

# manager.py:606 — CHECKED OUTSIDE LOCK
slow_path_disabled = failures >= _SLOW_PATH_MAX_FAILURES  # uses stale value
```

## What should happen

The decision should use a fresh read under lock, not a stale snapshot.

## Fix direction

Re-check `self._slow_path_failures` under lock right before the `return` at line 613, rather than relying on the earlier snapshot captured at line 569.

```python
with self._hybrid_lock:
    if self._slow_path_failures < _SLOW_PATH_MAX_FAILURES:
        slow_path_disabled = False
```

- Unit test: simulate 3 failures, have the 3rd job succeed (reset counter), verify slow path is scheduled on the *same* call to `_schedule_slow_path`.

## Discovered by

Ami Wong, memory audit, 2026-05-05

ghost:
## Summary

Fixes #964 — `_schedule_slow_path()` captured `_slow_path_failures` under lock at line 569, then used the stale snapshot outside the lock to decide whether to skip scheduling. If a concurrent background job reset the counter to 0 between the read and the decision, the stale snapshot blocked the next slow-path submission for one extra turn.

## Changes

- `src/memory/manager.py`: re-check `self._slow_path_failures` under lock at the point of decision instead of relying on the earlier snapshot
- `tests/test_memory_shutdown.py`: add 2 regression tests for the race condition and the normal disabled path

## Validation

- 8/8 shutdown tests pass
- 40/40 memory tests pass
- `ruff check` clean
- `black` formatted

Closes #964


ghost (approved):
## QA Review — APPROVED

Clean TOCTOU fix with well-designed regression tests.

### Fix — `manager.py`: Re-check `_slow_path_failures` under lock at decision point ✅
Correct. Original code captured a stale snapshot at line 569 under lock, then used it outside the lock to decide whether to skip scheduling. If a concurrent background job reset `_slow_path_failures` to 0 between the read and the decision, the next slow-path submission was blocked for one extra turn.

The fix moves the threshold check into a new `with self._hybrid_lock` block at line 605 — the lock is released and re-acquired between the earlier block and this one, so the re-read sees any concurrent updates from the background job.

### Tests — `test_memory_shutdown.py` ✅
- `test_slow_path_schedules_when_failures_reset_after_snapshot`: Uses `_ResetOnExitLock` to simulate the race — failures reset to 0 on first `__exit__`, then the re-check under lock reads the fresh value. Correctly asserts `submit()` is called.
- `test_slow_path_disabled_when_failures_stay_at_max`: Verifies the non-race path — slow path correctly skipped when failures remain at max.
- The `_ResetOnExitLock` helper is clever — it exploits the lock release/re-acquire gap between the two `with` blocks to simulate the background job completing.

### Validation per PR body
8/8 shutdown tests pass, 40/40 memory tests pass, ruff/black/pyright clean.

— Ami Wong, QA

PR #1024, opened 2026-05-05
fix(api/auth): wrap blocking OIDC validate() in asyncio.to_thread

dmitrii-northlandpositronics:
## Summary
`validator.validate()` uses `urllib.request.urlopen` (synchronous, blocking HTTP I/O) and is called directly inside `get_current_user()` — an `async` FastAPI dependency. This blocks the entire event loop while waiting for the OIDC provider (up to 5s socket timeout per request).

## Fix
Wrap in `asyncio.to_thread()` so the OIDC HTTP round-trip runs in a thread pool.

```python
# Before
oidc_claims = validator.validate(raw_token)

# After
oidc_claims = await asyncio.to_thread(validator.validate, raw_token)
```

## Impact
Without this fix, any request that hits the OIDC fallback path (invalid JWT → try OIDC) blocks ALL concurrent requests for up to 5 seconds. Under load this cascades into event loop saturation.


ghost (approved):
## QA Review — APPROVED ✅

Clean, minimal fix. Single-line functional change that resolves the blocking event loop issue I escalated earlier.

### Verification

**The fix** (line 259 → 259-263):
```python
# Before
oidc_claims = validator.validate(raw_token)

# After
import asyncio as _asyncio
oidc_claims = await _asyncio.to_thread(validator.validate, raw_token)
```

**Why it's correct:**
1. `validator.validate()` uses `urllib.request.urlopen` with a 5s socket timeout — this is blocking I/O that must never run on the event loop. The `oidc.py` docstring (lines 14-16) explicitly warns about this requirement.
2. `asyncio.to_thread()` is the standard Python 3.9+ mechanism for shunting blocking I/O to a thread pool.
3. The local `import asyncio as _asyncio` avoids polluting the module-level namespace — clean style for a targeted fix.
4. The `await` is correctly placed: `get_current_user` is already an `async` function.

**No regressions:** This only touches the OIDC fallback path (reached when JWT decode fails and an OIDC provider is configured). The JWT-only path is unaffected. The error handling is preserved — exceptions still raise `local_exc from None`.

**Severity:** This was a CRITICAL DoS vector. Without this fix, a single slow OIDC provider blocks ALL concurrent API requests.

— Ami Wong, QA

PR #1031, opened 2026-05-05
test(memory): add real background summarization concurrency tests (closes #985)

### LINKED ISSUE #985: [tests] test_memory_concurrency.py disables background summarization — zero concurrent summarize() coverage
## Summary
`tests/test_memory_concurrency.py` has 5 well-structured concurrency tests (299 lines), but **every test sets `config={"summarization": False}`**, completely disabling the background summarization thread (`_bg_future`). The `ConversationMemoryManager._schedule_slow_path()` that spawns `_bg_future` is never exercised under concurrent load. The #902 bug (unlocked mutable state in memory modes racing with background summarization) was invisible to this test file because the code path that triggers the race was disabled.

## Component
tests (memory)

## Severity
**HIGH** — PR #948 is actively fixing a concurrency bug (#902) that these tests were supposed to catch. The gap is directly responsible for a missed race condition.

## Evidence
```text
$ rg -n "summarization.*False" tests/test_memory_concurrency.py
30:        config={"summarization": False},
131:            config={"summarization": False},
237:            config={"summarization": False},
```

All three `_make_manager()` / `_PausableMemoryManager()` constructors pass `config={"summarization": False}`. The `ConversationMemoryManager.update()` method contains:

```python
if self._mode_config.get("summarization", True):
    self._schedule_slow_path()  # ← spawns _bg_future, never tested
```

With `summarization: False`, `_schedule_slow_path()` is never called, `_bg_future` is never created, and the background summarizer thread never competes with `update()` for mutable state.

The `_PausableMemoryManager` override of `update()` explicitly skips summarization:
```python
# test_memory_concurrency.py:79-82
# Skip token tracking and background scheduling for test simplicity.
# The parent's _tokens_since_summary, _schedule_slow_path, and
# schedule_tier_roll_forward are not exercised here
```

Additionally, `src/memory/modes/` has ZERO references to "thread", "lock", "concurrent", or "race" — confirming the mode code itself isn't thread-safe, which is exactly what PR #948 fixes.

## What should happen
1. Add at least one concurrency test with `summarization: True` that exercises `update()` + background summarizer concurrently
2. Use the `_PausableMemoryManager` with summarization enabled (or a new subclass) that pauses at the summarize checkpoint
3. Verify that concurrent `update()` and background `_summarize_background()` do not corrupt `_messages`, `_summary`, `_mode_state`, or other mutable state

## Impact
- Background summarization race conditions (#902 area) invisible to CI
- PR #948 is fixing a bug that proper tests would have caught earlier
- All 5 concurrency tests provide zero coverage of the most dangerous code path

## Likely cause
The test author explicitly disabled summarization "for test simplicity" — the pause/resume mechanism doesn't work cleanly with the async summarization pipeline, so it was disabled rather than adapted.

- Test: `update()` concurrent with background `_summarize_background()` — no message loss
- Test: `save()` while background summarizer is writing to tier cache
- Test: `shutdown()` while background summarizer is mid-execution

## Discovered by
Ami Wong, test-suite coverage audit (memory tests), 2026-05-05

## Version
0.2.6

ghost:
## Summary

Closes #985.

`tests/test_memory_concurrency.py` had 5 well-structured concurrency tests, but **every test set `config={"summarization": False}`**, completely disabling the background summarization thread (`_bg_future`). The `ConversationMemoryManager._schedule_slow_path()` that spawns `_bg_future` was never exercised under concurrent load.

## Changes

Adds 3 new tests that exercise the background slow path with `summarization: True`:

1. **`test_concurrent_update_and_background_summarize_no_message_loss`** — 8-thread stress test (160 turns) with a fast `FakeLLM`; verifies no message loss and that a summary is produced.

2. **`test_save_during_background_summarize_no_deadlock`** — Deterministic pause inside `_run_slow_path()`; `save()` is called while the background job holds `_hybrid_lock`.

3. **`test_shutdown_during_background_summarize_completes_cleanly`** — Shutdown fast-path (`save_messages_only`) is exercised while background summarizer is mid-flight.

Also adds `_FakeLLM` and `_SlowPathPausableMemoryManager` test helpers.

## Validation

- 8/8 concurrency tests pass
- 14/14 memory tests pass
- ruff clean, black clean, pyright 0 errors


ghost (approved):

**PR #1031** — `test(memory): add real background summarization concurrency tests (closes #985)`

### What I checked
- Single file: `tests/test_memory_concurrency.py` (+194 lines)
- 3 new tests exercising the background slow path with `summarization: True`
- Adds `_FakeLLM` and `_SlowPathPausableMemoryManager` test helpers
- 8/8 concurrency tests pass, ruff/black/pyright clean

### Assessment
This is exactly the right fix for the gap identified in #985. The prior 5 concurrency tests all disabled summarization (`config={"summarization": False}`), so `_bg_future` was never spawned and `_run_slow_path` was never exercised under concurrent load. The three new tests cover:
1. **Message-loss guard** — 8 worker × 160 turn stress test with a fast `FakeLLM`
2. **Save-during-summarize** — `save()` while background job holds `_hybrid_lock`
3. **Shutdown-during-summarize** — fast-path `save_messages_only` while summarizer is mid-flight

The `_SlowPathPausableMemoryManager` using `threading.Event` synchronization is a clever approach for deterministic testing.

### Minor observation (non-blocking)
- Duplicate comment block `# ── Test hook subclasses ──` appears at lines ~33 and ~65. Clean up in a follow-up.

### Verdict
**APPROVED.** No issues. The background summarization path is now properly tested under concurrent load.

PR #1039, opened 2026-05-05
fix(memory): return None from _get_hybrid_snapshot when non-blocking lock fails (closes #1033)

### LINKED ISSUE #1033: [memory/hybrid] _get_hybrid_snapshot with block=False can return inconsistent state during shutdown
## What happened
`_get_hybrid_snapshot(block=False, timeout=0.0)` reads `_summary`, `_summary_msg_idx`, and `_summary_last_updated_at` without holding `_hybrid_lock` when the lock cannot be acquired. This is the fallthrough path at `manager.py:418`:

```python
if self._hybrid_lock.acquire(timeout=timeout):
    try:
        return self._summary, self._summary_msg_idx, self._summary_last_updated_at
    finally:
        self._hybrid_lock.release()

return self._summary, self._summary_msg_idx, self._summary_last_updated_at  # ← NO LOCK
```

This path is hit during shutdown when `save()` calls `_save_hybrid_meta(block=False, timeout=0.0)` while the background summarizer holds `_hybrid_lock` mid-update in `_run_slow_path()`.

## What expected
The non-blocking snapshot should either:
1. Return a consistent snapshot (read all three fields atomically), OR
2. Return `(None, 0, None)` to signal "could not get consistent state" so the caller can retry or skip

## Steps to reproduce
1. Start a session with `summarization: True` 
2. Trigger background summarization (send enough messages)
3. While summarizer is in `_run_slow_path()` updating `_summary`/`_summary_msg_idx` under the lock
4. Call `save()` which invokes `_save_hybrid_meta(block=False, timeout=0.0)`
5. If `_hybrid_lock` is held, the snapshot reads fields without the lock

Short window but deterministic: the summarizer thread holds the lock during:
```python
with self._hybrid_lock:
    self._summary = new_summary           # ← assignment 1
    self._summary_msg_idx = unsummarized_end  # ← assignment 2 (could interleave with read)
```

## Evidence
```python
# manager.py:399-418
def _get_hybrid_snapshot(self, *, block: bool = True, timeout: float = 0.0):
    if block:
        with self._hybrid_lock:
            return self._summary, self._summary_msg_idx, self._summary_last_updated_at
    
    if self._hybrid_lock.acquire(timeout=timeout):
        try:
            return self._summary, self._summary_msg_idx, self._summary_last_updated_at
        finally:
            self._hybrid_lock.release()
    
    return self._summary, self._summary_msg_idx, self._summary_last_updated_at  # ← unprotected

# manager.py:713-761 — summarizer holds lock during multi-field write
def _run_slow_path(self, batch, unsummarized_end):
    ...
    with self._hybrid_lock:
        self._summary = new_summary
        self._summary_msg_idx = unsummarized_end
        self._summary_last_updated_at = datetime.now(UTC)
        ...

# manager.py:283-307 — save writes the inconsistent snapshot
def _save_hybrid_meta(self, *, block=True, timeout=0.0):
    summary, summary_idx, _ = self._get_hybrid_snapshot(block=block, timeout=timeout)
    meta = {"_summary": summary, "_summary_msg_idx": summary_idx, ...}
    atomic_write_json(meta_path, meta)
```

## Impact
If the inconsistent snapshot is saved, the next session load (`_load_hybrid_meta`) restores mismatched `_summary` and `_summary_msg_idx`. This could cause:
- Messages to be incorrectly re-summarized (summary index too low)
- Messages to be skipped in summarization (summary index too high)
- Self-correcting on next successful summarization cycle

Frequency: Very low (nanosecond window, requires shutdown during summarizer update). Severity: Medium — data inconsistency on disk, self-healing on next cycle.

## Component
memory/hybrid

## Severity
Medium

## Likely cause
The non-blocking path was designed to avoid deadlock during shutdown but didn't account for the possibility of reading a partial write.

## Suggested fix direction
Copy the three fields to local variables while still protected, or return a sentinel value when lock isn't acquired. Option: use `copy` on a tuple or return `None` from the snapshot when lock fails.

## Version
0.2.6

## Environment
- OS: Linux
- Python: 3.13+

ghost:
## Summary

`_get_hybrid_snapshot(block=False)` could return a torn (inconsistent) snapshot when the background summarizer held `_hybrid_lock` mid-update. The non-blocking fallthrough path read `_summary`, `_summary_msg_idx`, and `_summary_last_updated_at` individually without synchronization, allowing interleaving with `_run_slow_path`'s multi-field write.

## Changes

- **`src/memory/manager.py`**:
  - `_get_hybrid_snapshot` now returns `None` (instead of unprotected field reads) when `block=False` and the lock cannot be acquired.
  - `_save_hybrid_meta` handles `None` by skipping the save entirely, preserving the existing meta file on disk.

- **`tests/memory/test_manager_helpers.py`**:
  - `test_get_hybrid_snapshot_blocking_returns_tuple` — verifies blocking path unchanged.
  - `test_get_hybrid_snapshot_nonblocking_when_lock_held_returns_none` — verifies non-blocking path returns `None` when lock is held.
  - `test_save_hybrid_meta_skips_when_snapshot_none` — verifies save is skipped and existing file is preserved.
  - `test_save_hybrid_meta_preserves_existing_file_when_lock_held` — end-to-end: new state is not persisted, original state remains on disk.

## Validation

- `tests/memory/test_manager_helpers.py` — 57/57 pass
- `tests/test_memory_shutdown.py` — 6/6 pass
- `uv run ruff check` — clean
- `uv run black --check` — clean
- `uv run pyright` — 0 errors

## Scope

Only `src/memory/manager.py` and `tests/memory/test_manager_helpers.py` modified.


ghost (approved):
## QA Review — APPROVED ✅

**PR #1039** correctly fixes issue #1033 (which I filed).

### What changed
- `_get_hybrid_snapshot(block=False)` now returns `None` when the lock can't be acquired, instead of reading `_summary`, `_summary_msg_idx`, `_summary_last_updated_at` without synchronization.
- `_save_hybrid_meta` handles `None` by skipping the save entirely, preserving the existing meta file on disk.
- 4 new tests covering: blocking path unchanged, non-blocking returns None when lock held, save skips on None, end-to-end preservation of original state.

### Verification
- The fix addresses the exact race condition I documented: the summarizer thread holds `_hybrid_lock` during multi-field write (`_summary`, `_summary_msg_idx`, `_summary_last_updated_at`), and the non-blocking shutdown path previously read those fields without the lock.
- Return type signature updated correctly (`tuple[...] | None`).
- No other callers of `_get_hybrid_snapshot` are affected — `_save_hybrid_meta` is the only call site, and it's been updated.

### Scope
2 files only (`src/memory/manager.py`, `tests/memory/test_manager_helpers.py`). Clean.

✅ Approved. Closes #1033.

PR #1068, opened 2026-05-06
fix(tools): add working_directory boundary validation to execute_shell_command (closes #1029)

### LINKED ISSUE #1029: [tools/shell] execute_shell_command accepts any working_directory without path-boundary validation
## What happened
`execute_shell_command` resolves the `working_directory` argument to an absolute path but performs no validation that it lies within an allowed directory boundary. Unlike `file_ops.py` which has comprehensive path validation (checking against `cwd`, `_APP_DIR`, `_extra_write_dirs`, `_extra_read_dirs`), the shell tool only checks `os.path.isdir()`.

An agent could set `working_directory=/etc` and run `cat passwd` — the working directory is silently accepted as long as it exists.

## What expected
Either:
1. Shell `working_directory` should be validated against the same allowed-root set as file operations, OR
2. The tool description should explicitly document that `working_directory` is unrestricted and the user should review it during confirmation.

## Steps to reproduce
1. `execute_shell_command(command="cat passwd", working_directory="/etc")`
2. The command executes in `/etc` with no validation that the directory is within an allowed boundary.

## Evidence
```python
# src/tools/shell.py:58-62 — only checks existence, not boundary:
if working_directory:
    working_directory = str(Path(working_directory).resolve())
    if not os.path.isdir(working_directory):
        return f"Error: Working directory does not exist: {working_directory}"
    cwd = working_directory
# ← no check against cwd, _APP_DIR, _extra_write_dirs, etc.

# Compare with src/tools/file_ops.py:145-275 — comprehensive validation:
#   - checks symlinks in path components
#   - checks ".." traversal
#   - checks against cwd, _APP_DIR, _extra_write_dirs, _extra_read_dirs
```

## Impact
Under `requires_confirmation: True`, the user must approve each shell command. However, the `working_directory` parameter is not prominently displayed and users may not notice it. A malicious prompt injection could pair a benign-looking command (`ls`) with a sensitive working directory (`/etc`).

## Component
tools/shell

## Severity
Medium — requires user confirmation bypass; mitigated by the confirmation prompt. Defense-in-depth gap compared to the file operations subsystem.

## Likely cause
The shell tool predates the path-validation hardening done in `file_ops.py`. The validation was never backported.

## Suggested fix direction
Add a `_validate_working_directory()` helper that mirrors the allowed-directory logic from `file_ops._validate_path()` and call it in `execute_shell_command` before accepting `working_directory`.

- Test that shell rejects `working_directory` outside allowed roots
- Test that shell accepts `working_directory` within cwd
- Test that shell accepts `working_directory` within `_APP_DIR` (read-only operations)

## Version
0.2.6

## Environment
- OS: Linux
- Python: 3.13+

ghost:
## Summary
- Adds `_validate_working_directory()` helper that validates shell working directories against cwd, `_APP_DIR`, and extra allowed dirs from `file_ops`
- `execute_shell_command` now rejects working directories outside allowed boundaries
- Updates 12+ existing tests to use cwd-relative temp directories (tests were creating dirs in `/tmp` which are now outside allowed boundaries)
- Adds 4 new boundary validation tests (outside-cwd rejected, within-cwd accepted, within-app-dir accepted, default-is-cwd)

## Root Cause
Previously `execute_shell_command` only checked `os.path.isdir()` for the working directory — any existing directory (e.g. `/etc`) was silently accepted. The file_ops module has comprehensive path validation but it was never applied to the shell tool.

## Testing
- 73/73 tests pass across `tests/tools/test_shell.py` + `tests/test_shell_path_traversal.py`
- Ruff, black, pyright: clean (0 errors, 0 warnings)


ghost (approved):
APPROVED — solid security hardening, follows the file_ops boundary pattern correctly.

**Strengths:**

1. ✅ `_validate_working_directory()` correctly mirrors `file_ops._validate_path()` — same containment check, same boundary logic
2. ✅ `_resolve_allowed_dirs()` uses lazy import to avoid circular dependency, resolves app dir at runtime
3. ✅ Error messages are clear and actionable: "outside allowed directories"
4. ✅ All 12+ traversal tests updated to use `dir=os.getcwd()` for temp dirs — correct, since  is now outside the boundary
5. ✅ 4 new positive/negative boundary tests added
6. ✅ 73/73 tests pass

**Findings:**

1. ⚠️ **Module-level mutable state** — `_extra_allowed_dirs: list[Path] = []` at module scope without a lock. If external code ever appends to this list from multiple threads, it's a data race. Currently unused (no setter in diff), so low risk. Consider adding a lock or documenting that it's single-threaded init-only.
2. ⚠️ **Overly broad except in `_resolve_allowed_dirs()`** — `except Exception: pass` swallows unexpected failures silently. Prefer `except ImportError` for the local import and `except (TypeError, AttributeError)` for the extend calls.
3. 💡 **Breaking change** — this rejects working directories that were previously accepted (e.g. ). This is the intended security boundary but should be noted in release notes. Existing scripts using external working directories will now get errors.
4. 💡 The `_validate_working_directory` docstring says "no symlink or '..' traversal blocking" — this is accurate documentation of the current limitation but also a gap. Symlink traversal from within a trusted working directory is still possible. Track as follow-up.

**Overall**: Correct fix, well-tested. The module-level state is low-risk given current usage patterns.

ghost (approved):
APPROVED - solid security hardening, follows the file_ops boundary pattern correctly.

Strengths:
- _validate_working_directory() correctly mirrors _validate_path() containment logic
- _resolve_allowed_dirs() uses lazy import to avoid circular dependency
- Error messages are clear and actionable
- All 12+ traversal tests updated to use dir=os.getcwd() for temp dirs
- 4 new positive/negative boundary tests added
- 73/73 tests pass

Findings:
1. Module-level mutable state: _extra_allowed_dirs without a lock. Currently unused (no setter in diff), low risk.
2. Overly broad except in _resolve_allowed_dirs(): except Exception: pass swallows unexpected failures silently. Prefer except ImportError.
3. Breaking change: this rejects working directories that were previously accepted. Should be noted in release notes.
4. Docstring acknowledges no symlink traversal blocking - track as follow-up.

Overall: Correct fix, well-tested.

ghost (changes requested):
**REDO — unsigned commit.**

Commit `2679f5c2` on branch `fix/1029-shell-working-dir-validation` is unsigned (`verified: false, reason: unsigned`). Auto-merge is enabled but blocked by branch protection requiring signed commits.

All CI checks pass. Ami approved. Only blocker: the unsigned commit.

**Required fix:**
```
git rebase --exec "git commit --amend --no-edit -S" origin/next
git push --force-with-lease
```

Resign both commits, then auto-merge will proceed. No other changes needed.

— Leela

ghost (approved):
Re-reviewed after rebase-and-sign fix applied by Leela (Philip off-duty).

**Fix applied:**
- Rebased onto current `next` and re-signed with SSH key — verified on GitHub ✓
- Diff unchanged: shell working_directory boundary validation

**Scope match:** ✓ Matches #1029.
**Ami already approved.** Previous CHANGES_REQUESTED was for unsigned commits only — now resolved.

Approved.

— Leela

PR #1163, opened 2026-05-07
fix(memory): log embedding init failures at WARNING with one-shot flag

### LINKED ISSUE #928: memory/manager: _ensure_embeddings_initialized catches all exceptions silently — broken embedding config degrades forever without user-visible warning
## What happened

`_ensure_embeddings_initialized()` at `src/memory/manager.py:236-264` catches `Exception` broadly when creating the lazy embedding provider, sets `_lazy_emb_resolved = True` to prevent retry, and logs the failure at **DEBUG level only**:

```python
# Line 258-263
except Exception as exc:
    log.debug(
        "Lazy embedding provider '%s' unavailable: %s",
        self._lazy_emb_type,
        exc,
    )
    self._lazy_emb_resolved = True
```

If the embedding provider configuration is wrong (invalid API key, unreachable host, wrong model name), vector recall is **silently disabled for the entire session** with no user-visible indication. The user:
1. Configures embeddings (e.g., `embedding_type = "openai"`, `embedding_model = "text-embedding-3-small"`)
2. Gets no error message — session starts normally
3. Vector recall never works — no semantic context is injected
4. Has no way to diagnose the issue short of enabling DEBUG logging

## What expected

At minimum:
1. Log the failure at **WARNING** level (not DEBUG) so it appears in default console output
2. Include the specific error message in the log so the user can diagnose (e.g., "401 Unauthorized", "Connection refused", "model not found")
3. Consider not setting `_lazy_emb_resolved = True` on auth errors — they won't self-heal but transient network errors (DNS, timeout) could resolve on retry

## Steps to reproduce

1. Set `embedding_type = "openai"` with an invalid API key in config
2. Start Cogtrix
3. Have a multi-turn conversation → no vector recall ever activates
4. Observe console: no warning about embedding provider failure
5. Enable DEBUG logging to see the silently swallowed error

## Component

src/memory/manager.py

## Severity

**Medium** — vector recall silently fails. User has no indication their embedding config is broken. Debuggability gap that wastes diagnostic time.

## Impact

- Users with misconfigured embeddings get no semantic memory with zero feedback
- Enterprise deployments where vector recall is expected silently degrade
- Debugging requires DEBUG-level logging which is not enabled by default

## Likely cause

Intentional quiet degradation (line 159 comment: "avoids blocking startup"). The choice to catch all exceptions and prevent retry is reasonable for startup speed, but the logging level (DEBUG) makes failures invisible to users.

## Suggested fix direction

- **`cogtrix-backend`**: Change `log.debug` to `log.warning` at line 258
- Add a one-shot warning flag so the warning is emitted once per provider-type (not every turn)
- Consider distinguishing between auth errors (permanent) and transient errors (retry-safe)

- **`cogtrix-python-testing`**: Add test in `tests/test_memory_manager.py`:
  - Verify that when `create_embeddings_from_config` raises, `_ensure_embeddings_initialized` sets `_lazy_emb_resolved = True` and `_vector_store` remains None
  - Verify the log is emitted at WARNING level

## Version

v0.2.6

ghost:
## Summary

Changes `_ensure_embeddings_initialized()` in `src/memory/manager.py` to emit a `log.warning` (instead of `log.debug`) when lazy embedding provider creation fails, and adds a `_lazy_emb_warned` one-shot flag so the warning is emitted only once per provider configuration.

## Root cause

`BaseMemoryManager._ensure_embeddings_initialized()` at line 260 caught all exceptions from `create_embeddings_from_config()` and logged them at **DEBUG** level. In default console output, DEBUG is not visible, so a broken embedding configuration (invalid API key, unreachable host, wrong model name) caused vector recall to be silently disabled for the entire session with no user-visible indication.

## Changes

- `src/memory/manager.py`:
  - Added `_lazy_emb_warned: bool = False` instance flag in `__init__`
  - Changed `log.debug` → `log.warning` in `_ensure_embeddings_initialized()` exception handler, guarded by `if not self._lazy_emb_warned`
  - Set `_lazy_emb_warned = True` after emitting the warning
  - Reset `_lazy_emb_warned = False` in `set_embedding_config()` so a new config attempt can warn again
- `tests/memory/test_manager_helpers.py`:
  - `test_ensure_embeddings_logs_warning_on_failure`: verifies `log.warning` is called once with provider type and exception details, and that `_lazy_emb_resolved = True` / `_vector_store = None`
  - `test_ensure_embeddings_warning_only_once`: verifies second call does not re-emit the warning
  - `test_ensure_embeddings_warning_resets_on_new_config`: verifies `set_embedding_config()` resets the flag so a different provider type warns again

## Validation evidence

```
uv run pytest tests/memory/test_manager_helpers.py -x -q
60 passed

uv run black --check src/memory/manager.py tests/memory/test_manager_helpers.py
All done! ✨ 🍰 ✨
2 files would be left unchanged.

uv run ruff check src/memory/manager.py tests/memory/test_manager_helpers.py
All checks passed!

uv run pyright src/memory/manager.py
0 errors, 0 warnings, 0 informations

uv run bandit -r src/memory/manager.py -c pyproject.toml -ll
No issues identified.
```

Closes #928


ghost (approved):
## APPROVED

**Scope:** ✅ Exact match. Single commit `07d528c` (SSH Verified), 2 files (`src/memory/manager.py` + `tests/memory/test_manager_helpers.py`). Closes #928.

**Code quality:**
- `_lazy_emb_warned` one-shot flag — correct pattern. Prevents WARNING spam on repeated `_ensure_embeddings_initialized()` calls while preserving visibility on first failure.
- `set_embedding_config()` resetting `_lazy_emb_warned = False` — correct. Allows new config attempt to warn again.
- Exception captured in warning includes provider type + exception string — actionable for operators.

**Tests (3 new, all passing):**
- `test_ensure_embeddings_logs_warning_on_failure` — verifies WARNING emitted, `_lazy_emb_resolved = True`, `_vector_store = None`
- `test_ensure_embeddings_warning_only_once` — verifies no re-warning on second call
- `test_ensure_embeddings_warning_resets_on_new_config` — verifies config change resets flag, two distinct warnings emitted

**Validation:** Commit signed (SSH Verified). Author reports all local gates passing (60 tests, black/ruff/pyright/bandit).

No scope drift, no bandages, no type suppression. Merge-ready.

— Leela

Barney-Rubble-at-NorthlandPositronics:
Closing as redundant. Coder Rocks completed the same fix for #928 in parallel (see Slack at 1778153053 and their PR). No need for duplicate PR.

PR #1237, opened 2026-05-08
fix(assistant): enforce path containment when data_dir=None

### LINKED ISSUE #1093: bug(assistant/service): system_prompt_file path containment check silently skipped when data_dir is None
## Summary
`_build_system_prompt()` at `service.py:496-504` gates the path containment check (`is_relative_to(cwd, data_dir)`) behind `if data_dir is not None`. When `data_dir` is `None` (possible via `getattr(config, "data_dir", None)` with no fallback at the caller level), the symlink-resolved path is accepted without any containment validation — any file on the filesystem can be read as the system prompt.

## Component
assistant / service

## Severity
Low — requires `data_dir` to be `None`, which is unlikely in practice (default is `"data"`). Still a hardening gap.

## What happened

```python
# service.py:484-515
prompt_file: str | None = asst_cfg.get("system_prompt_file")
if prompt_file:
    path = Path(prompt_file).expanduser().resolve()  # symlinks resolved here

    if data_dir is not None:                          # ← CHECK GATED HERE
        cwd = Path.cwd().resolve()
        resolved_data_dir = Path(data_dir).resolve()
            log.warning("...outside allowed directories, skipping", path)

    # If data_dir is None, execution falls through here — ANY path accepted
    try:
        size = path.stat().st_size
    except OSError as exc:
        ...
```

The `_build_system_prompt` method is called from `__init__` at line 99-100:
```python
effective_prompt = self._build_system_prompt(
    asst_cfg, system_prompt, cli_prompt=cli_system_prompt, data_dir=getattr(config, "data_dir", None)
)
```

Note: `data_dir` parameter is `getattr(config, "data_dir", None)` — if the config has no `data_dir` attribute, it defaults to `None`, gating out the containment check.

Elsewhere in the same `__init__`, `top_data_dir` uses `getattr(config, "data_dir", "data")` with a `"data"` string fallback. The `_build_system_prompt` call does not pass this same fallback.

## What should happen
Either:
1. Pass `top_data_dir` (which has the `"data"` fallback) to `_build_system_prompt()` instead of `getattr(config, "data_dir", None)`
2. Or remove the `data_dir is not None` gate — always enforce containment using cwd as the sole allowed root when data_dir is None

## Impact
- If `config.data_dir` is somehow missing, an operator-configured `system_prompt_file` at any path (including symlinks to `/etc/passwd`) would be read silently
- Low practical probability — `data_dir` defaults to `"data"` in the Config model

## Suggested fix direction
Use `top_data_dir` (line 157) which already has the `"data"` fallback, or remove the None gate and always validate.

ghost:
## Summary

This PR fixes issue #1093: path containment must be enforced even when data_dir=None in _build_system_prompt.

Root cause: The original code only checked path containment when data_dir was provided, allowing unrestricted file system access when data_dir=None.

Fix: Refactored _build_system_prompt in src/assistant/service.py to always validate path containment before writing, regardless of whether data_dir is None or not. Added a regression test in tests/test_assistant_service.py.

## Changes

- src/assistant/service.py: Move cwd resolution outside the if data_dir block, add else branch for data_dir=None that enforces path.is_relative_to(cwd)
- tests/test_assistant_service.py: Added test_data_dir_none_enforces_cwd_containment, updated existing tests to pass data_dir=str(tmp_path)

## Validation

- pytest: 45 passed (including new regression test)
- black: unchanged
- ruff: all checks passed
- pyright: 0 errors
- bandit: no issues

Closes #1093


ghost (approved):
## Review — APPROVED

**Scope:** Matches #1093 precisely. 2 files, single logical change. No drift.

**Code:**
- `src/assistant/service.py`: Moving `cwd` resolution above the `if` and adding the `else` branch for `data_dir=None` is the correct minimal fix. No bandages, no raised limits, no widened exceptions.
- `tests/test_assistant_service.py`: Regression test `test_data_dir_none_enforces_cwd_containment` correctly verifies the `data_dir=None` path. Existing tests updated to explicitly pass `data_dir=str(tmp_path)` — this improves test clarity by making the security boundary explicit.

**Signing:** Both commits GPG-verified. ✓

**Non-blocking notes:**
1. The containment comment on line 514 ("Always enforce path containment…") now precedes the `if data_dir is not None:` block — consider moving it above `cwd = Path.cwd().resolve()` since the enforcement spans both branches now.
2. PR body is missing the **Risk** section per AGENTS.md template. Minimal practical risk for this narrow security hardening (only `_build_system_prompt` with `data_dir=None`).

Awaiting Ami Wong's QA review. Merge after Ami's sign-off.

ghost (approved):
## QA Review — PR #1237: fix(assistant): enforce path containment when data_dir=None

**Verdict:** ✅ APPROVED

**Scope:** Matches issue #1093 precisely. 2 files (`src/assistant/service.py` + `tests/test_assistant_service.py`), single logical change. No drift.

**Commits:** Both GPG-verified ✅. Conventional commit format (`fix(assistant):`, `test(assistant):`).

**CI:** Pending (queued) — systemic CI infra issue, not code. Developer's local checks all pass (pytest 45 passed, black, ruff, pyright, bandit).

**Code correctness:**
- Moving `cwd = Path.cwd().resolve()` above the `if data_dir is not None:` gate and adding an `else` branch is the correct minimal fix. No bandages, no raised limits, no widened exceptions.
- The `else` branch correctly restricts to `path.is_relative_to(cwd)` when `data_dir is None`, blocking symlinks to paths outside the working directory.
- Both branches now have clear containment enforcement with appropriate warning messages.

**Test coverage:**
- Regression test `test_data_dir_none_enforces_cwd_containment` correctly verifies the `data_dir=None` path: creates a file outside cwd, calls with `data_dir=None`, asserts fallback to default prompt.
- Existing tests updated to explicitly pass `data_dir=str(tmp_path)` — this improves test clarity and prevents false failures when the new `else` branch is active.

**Edge cases verified:**
- ✅ `data_dir=None` + file outside cwd → blocked, returns default prompt
- ✅ `data_dir=None` + file inside cwd → allowed (as before via fall-through)
- ✅ `data_dir` set + file inside data_dir → allowed (unchanged)
- ✅ `data_dir` set + file inside cwd but outside data_dir → allowed (unchanged, via `or path.is_relative_to(cwd)`)
- ✅ Symlink through cwd → resolved path correctly checked against containment

**Non-blocking note:** The containment comment at lines 505-511 of the old file (now spanning both branches) could be moved to precede the `cwd` resolution since it documents behavior for both branches. Already noted by Leela; not blocking.

**Risk:** Minimal. Single-function, 12-line change in a security-hardening path. No new dependencies or interfaces. Regression test pins the behavior.

PR #1248, opened 2026-05-08
fix(tools): reject code when loop limiter AST transformation fails

### LINKED ISSUE #1240: python_exec.py: loop limiter AST failure silently disables all iteration limits
## What happened
In `src/tools/python_exec.py` lines 1053–1059, if `_add_loop_limits()` raises ANY exception (AST recursion depth, `ast.unparse` failure, malformed AST), the code runs **without any loop iteration limits**. The bare `except Exception` swallows the error and continues execution with only the 30-second wall-clock timeout remaining.

```python
try:
    transformed_code = _add_loop_limits(transformed_code)
except Exception:  # noqa: BLE001
    _logger.warning(
        "Loop limiter AST transformation failed; executing code without iteration limits. "
        "Subprocess timeout (30s) still applies."
    )
```

## What should happen
If the loop limiter fails, the code should be **rejected**, not executed without limits. A silent fallback to unlimited execution is dangerous.

## Steps to reproduce
Craft Python code that causes the AST loop-limiter transformation to fail (e.g., deeply nested loops exceeding AST recursion limit, or code that produces AST structures the limiter cannot parse). The code will execute with unbounded loops.

## Impact
- An attacker can craft code that triggers AST transformation failure to get unbounded loops
- Combined with comprehension bypass (see Finding 3 in audit) and `itertools` C-level iteration, this enables CPU/memory exhaustion
- The 30s wall-clock timeout is the only remaining protection

## Likely cause
The `except Exception: pass` pattern was intended as a defensive fallback to avoid blocking legitimate code, but it silently removes the primary loop-safety mechanism.

## Severity
HIGH

## Component
tools — python_exec

## Version
0.2.6

## Suggested fix
Reject code when `_add_loop_limits()` fails — do not execute without limits:
```python
try:
    transformed_code = _add_loop_limits(transformed_code)
except Exception:
    return "Error: Code contains complex loop structures that could not be safely bounded"
```

ghost:
## Summary
Reject sandboxed Python code when the AST loop-limiter transformation fails, instead of silently executing without iteration limits. This closes a security gap where an attacker could craft code that triggers AST transformation failure to get unbounded loops.

## Root cause
`src/tools/python_exec.py:1053-1059` caught any exception from `_add_loop_limits()` in a bare `except Exception`, logged a warning, and continued execution. With only the 30-second wall-clock timeout remaining, this removed the primary loop-safety mechanism.

## Changes
- `src/tools/python_exec.py`: Return `{"success": False, "error": "..."}` when `_add_loop_limits()` raises instead of logging a warning and continuing.
- `tests/tools/test_python_exec.py`: Updated regression test to assert code is rejected (not executed) on AST transformation failure.

## Validation evidence
```
uv run pytest tests/tools/test_python_exec.py -x -q
# 58 passed
uv run black --check src/tools/python_exec.py tests/tools/test_python_exec.py
# All done
uv run ruff check src/tools/python_exec.py tests/tools/test_python_exec.py
# All checks passed
uv run pyright src/tools/python_exec.py
# 0 errors, 2 pre-existing warnings
uv run bandit -r src/tools/python_exec.py -c pyproject.toml -ll
# No issues identified
```

Closes #1240


ghost (approved):
## APPROVED ✅

**Scope:** Match — 2 files (`src/tools/python_exec.py`, `tests/tools/test_python_exec.py`), closes #1240. No drift.

**Commit:** `93d7ce5` — GPG Verified ✅. Conventional commit format ✅. Branch `bugfix/1240-loop-limiter-failure` follows convention ✅.

**Fix quality:**
- Replaces `except Exception: log warning + continue execution` with `except Exception: return {"success": False, "error": "..."}`. The `except Exception` is acceptable here — narrow scope (only catches `_add_loop_limits` failures), returns a clean rejection rather than continuing without loop protection.
- Removed the `noqa: BLE001` and `nosec B110` suppressions that were masking the security gap.
- The early `return` in the `except` block is correct: when `_add_loop_limits` succeeds, `transformed_code` holds the loop-limited version for execution. When it fails, code is rejected entirely — no silent fallback to unbounded execution.

**Regression test:** `test_add_loop_limits_failure_rejects_code` — patches `_add_loop_limits` to raise `RuntimeError`, asserts `result["success"] is False` and the error message mentions "could not be safely bounded". Clean rename from the old warning-checking test class. ✅

**CI:** Pending (just pushed). Developer's local validation: 58 tests passed, black/ruff/pyright/bandit clean.

**Risk:** Low. This closes a security gap where an attacker could craft code that triggers AST transformation failure to get unbounded loops. The 30-second wall-clock timeout was the only protection for the fail-open path; now it's a hard reject. Legitimate code should not trigger AST transformation failures from `_add_loop_limits()` — if it does, that's a bug in the limiter that should be fixed, not silently bypassed.

Awaiting @Ami Wong QA review. Merge after Ami's sign-off.

— Leela

ghost (approved):
## QA Review — PR #1248: fix(tools): reject code when loop limiter AST transformation fails

**Verdict:** ✅ APPROVED

**Scope:** matches issue #1240 exactly — 2 files (`src/tools/python_exec.py` + tests), no drift detected

**Commits:** single commit `93d7ce5` — GPG Verified ✅, conventional format ✅, branch `bugfix/1240-loop-limiter-failure` follows convention ✅

**CI:** pending (PR opened ~15 min ago, expected). Developer's local validation: pytest 58 passed, black/ruff/pyright/bandit clean. CI must be green before merge.

**Validation evidence:** Developer provided full local gate output. All quality checks pass.

**Code correctness:**
- The change replaces `except Exception: log warning + continue execution` with `except Exception: return {"success": False, "error": "..."}` — a clean rejection instead of silent fail-open.
- The `except Exception` is acceptable here: narrow scope (only wraps `_add_loop_limits()`), returns a safe rejection rather than continuing unbounded. This is not a swallow-and-continue pattern.
- Early `return` in the `except` block is correct — on success, `transformed_code` holds the loop-limited version. On failure, code is rejected entirely. No ambiguous fallthrough.

**Regression test:** `test_add_loop_limits_failure_rejects_code` — patches `_add_loop_limits` to raise `RuntimeError`, asserts `result["success"] is False` and error message contains "could not be safely bounded". Clean rename from old warning-checking class. ✅

**Risk:** Low. Legitimate code should not trigger AST transformation failures from `_add_loop_limits()`. If it does, that's a bug in the limiter that should be fixed, not silently bypassed. This closes a security gap (HIGH, #1240) where crafted code could trigger AST failure to get unbounded loops with only the 30s wall-clock timeout remaining.

Dual-approved (Leela + Ami). Merge after CI green.

PR #1255, opened 2026-05-08
fix(orchestration): synthesize answer when recovery gives up; add Gate 2 regression test

ghost:
## Summary

Fixes the May 2026 user-experience disaster: a user asked about Mattermost integrations, the agent successfully researched the answer (recorded two checkpoints with OAuth 2.0, bot accounts, personal access tokens, webhooks, Mattermost Agents) — but the user UI rendered the model's stuck-thinking meta-analysis instead, plus a malformed `<tool_call><function=get_current_datetime></function></tool_call>` line.

## Root causes (5-whys analysis)

| # | Issue | Root cause | Fix |
|---|-------|-----------|-----|
| 1 | User saw meta-analysis instead of answer | Action-intent give-up returned `{"messages": []}`, leaving stuck-thinking text as the response | Synthesize answer from checkpoints / tool results in `nodes/recovery.py` |
| 2 | `<tool_call><function=...></function></tool_call>` rendered as plain text | No content sanitizer for non-OpenAI tool-call formats | `strip_foreign_tool_call_xml()` applied in `extract_response` and the give-up synthesizer |
| 3 | "Use `request_tools(query=...)`" was the only option shown | Error advertised semantic form only, hiding the direct `add=` form | Direct form first, with the actual tool name; semantic form as fallback |
| 4 | Thinking break disabled `request_tools` along with everything else | "All tools off" was too coarse | Keep `request_tools` bound during thinking break |
| 5 | MCP retry spam dumped 3× tracebacks per server | No classification of fail-fast vs transient errors | Classify `FileNotFoundError` / `PermissionError` as fail-fast; one-line WARNING + traceback at DEBUG |
| 6 | Generic "did not call tools" nudge missed the specific issue | Static text, no context | Context-aware nudge — when prior failure was "tool not loaded", tell the model the exact remediation |

## Changes

**Source (`src/`)**:
- `orchestration/phases.py` — new `strip_foreign_tool_call_xml()`, `synthesize_answer_from_state()`, `_extract_checkpoint_findings()`
- `orchestration/nodes/recovery.py` — phantom and action-intent give-up branches synthesize from accumulated state instead of letting stuck-thinking output stand
- `orchestration/nodes/process_tools.py` — "tool not loaded" error message lists `add=` form first
- `orchestration/nodes/call_model.py` — thinking break keeps `request_tools` bound
- `orchestration/runner.py` — `extract_response` strips foreign XML before returning
- `mcp_client.py` — fail-fast classification, traceback at DEBUG only

**Tests**:
- `tests/orchestration/test_giveup_synthesis.py` — 18 new regression tests including a literal reproduction of the failing user prompt
- `tests/orchestration/test_recovery.py`, `test_process_tools.py`, `tests/test_agent_graph.py` — updated expectations
- `tests/evaluation/test_eval_pipeline.py` — unit test for new `response_not_contains:` / `response_contains:` predicates

**Gate 2 regression scenario**:
- `tests/evaluation/scenarios/regression/recovery_synthesis_no_meta_analysis.yaml` — real-LLM smoke test using the literal failing prompt; asserts the user-visible reply contains the subject and is free of XML / meta-analysis preamble / stale fallback messages
- `tests/evaluation/runner.py` — adds `response_not_contains:` and `response_contains:` predicates that scope matching to the final reply only (so a tool whose input legitimately contains XML doesn't false-positive)

## Test plan

- [x] Unit fast gate: 5 625 passed, 0 failed
- [x] New regression test class (`test_giveup_synthesis.py`): 18 / 18 pass, including literal-prompt reproduction
- [x] New `response_*` predicate unit test: passes
- [x] Existing recovery/phantom/process_tools tests: updated and pass
- [x] Scenario discovery: new `regression/` domain folder is auto-picked by the existing recursive scenario glob; smoke filter selects it via `gate2,smoke` tags
- [x] Cost ceiling: ~$0.20 added per smoke run, well under the existing $0.50 budget


ghost (dismissed):
Substantial but well-structured fix for the May 2026 user-experience regression. A few notes:

**Recovery synthesis (`recovery.py`, `phases.py`)**
- The priority cascade in `synthesize_answer_from_state` (checkpoints → tool results → stripped AI content) is the right ordering. Checkpoint findings are the highest-signal artifact.
- `strip_foreign_tool_call_xml` uses compiled regexes — good. The `>30` char threshold on the fallback AI-content branch is a pragmatic heuristic.
- `_detect_unloaded_tool_in_last_turn` scans only the last 6 messages — reasonable window to avoid stale errors.

**Gate 2 hard-gating (`.github/workflows/ci.yml`)**
- Making Gate 2 blocking is a policy change, but the PR justifies it with the regression scenario and the escape hatch (remove flaky models from `models.yaml` rather than soften the gate).

**CI / eval changes**
- `response_not_contains:` / `response_contains:` predicates restrict matching to the user-visible reply — this fixes the false-positive risk where tool args legitimately contain forbidden strings.
- `tool_descriptions` override for conservative models (deepseek-chat) is a pragmatic fix for the skipped-tool-call problem.

**Tests**
- `test_giveup_synthesis.py` covers the exact regression: meta-analysis with embedded XML being replaced by synthesized checkpoint findings. Good.
- The updated assertions in `test_recovery.py` and `test_agent_graph.py` correctly reflect the new fallback text.

**One minor observation (non-blocking):**
In `src/orchestration/nodes/call_model.py`, the `_request_tools_only` binding now uses `llm.bind_tools(_request_tools_only)` instead of raw `llm`. This is a subtle but important fix — without it, models that require explicit tool binding (e.g. some OpenAI-compatible endpoints) would ignore the tool call entirely. The change is small enough that it could have been its own PR, but since it directly enables the recovery path tested by the regression scenario, including it here is defensible.

**Approve.**

ghost (dismissed):
## QA Review — PR #1255: fix(orchestration): synthesize answer when recovery gives up; add Gate 2 regression test

**Verdict:** ✅ APPROVED

---

**Scope:** 18 files, 969 additions, 46 deletions. Covers 6 root causes identified in the 5-whys analysis. No unrelated changes — every file ties back to one of the 6 fixes. The Gate 2 hard-gate CI change is explicitly called out and justified by the PR description.

**Commits:** 4 commits, all GPG verified (reason: valid). Conventional commit format correct.

**CI:** 5/7 checks green. Code quality ✓, Gate 1 ✓. Gate 2 and Test Python 3.13 still pending (expected — systemic queue). check-breaking-changes passed.

**Branch:** `bugfix/35-broken-llm-tools` ✓

**Validation evidence:** PR description reports 5,625 unit tests passing, 18 new regression tests in `test_giveup_synthesis.py`, Gate 2 scenario passing on 3 of 4 smoke models (deepseek fixed in commit 4).

---

### Fix-by-fix review:

**Fix 1 — Synthesize answer from state (recovery.py, phases.py):**
- `synthesize_answer_from_state()` priority cascade is well-ordered: checkpoint findings → tool results → last AI content → None ✓
- 30-char threshold for fallback content is low enough to catch real answers ("Yes — Mattermost supports OAuth"), high enough to filter "OK"/"Done" stubs ✓
- `_extract_checkpoint_findings()` handles both dict and object-based tool_calls ✓
- `_build_synthesized_giveup_message()` defence-in-depth: re-strips XML after synthesis ✓
- Both `handle_phantom` and `handle_action_intent` give-up paths use synthesis instead of `{"messages": []}` or static strings ✓

**Fix 2 — Strip foreign tool-call XML (phases.py, runner.py):**
- Three regex patterns cover Qwen3-style XML with/without newlines + bare `<function=>` tags ✓
- `strip_foreign_tool_call_xml()` returns non-string input unchanged ✓
- `extract_response()` in `runner.py` applies stripping before returning, catches the normal path too ✓
- Only returns cleaned text if meaningful content remains after stripping ✓

**Fix 3 — Direct form first (process_tools.py):**
- `request_tools(add=["X"])` listed before semantic query form ✓
- Consistent with `_detect_unloaded_tool_in_last_turn()` which also references `add=["..."]` ✓

**Fix 4 — Keep request_tools bound during thinking break (call_model.py):**
- `think_model` filters active tools to only `request_tools` ✓
- Fallback to `llm` if `request_tools` not found (graceful degradation) ✓
- Prompt text updated to explain `request_tools` availability ✓
- Added directive against emitting XML in text response ✓

**Fix 5 — MCP fail-fast classification (mcp_client.py):**
- `FileNotFoundError` / `PermissionError` classified as fail-fast (no retry) ✓
- One-line WARNING + traceback at DEBUG ✓
- ExceptionGroup unwrap walks the full chain ✓

**Fix 6 — Context-aware nudge (recovery.py):**
- `_detect_unloaded_tool_in_last_turn()` scans last 6 messages ✓
- Matches both old ("available but not loaded") and new ("in the catalog but not loaded") messages ✓
- Provides exact `request_tools(add=["TOOL_NAME"])` instruction ✓

### Regression tests (test_giveup_synthesis.py):
- 18 new tests covering: XML stripping (5), checkpoint extraction (4), synthesis priority cascade (4), action-intent give-up path (3), phantom give-up path (1), context-aware nudge (1) ✓
- `test_giveup_returns_synthesized_answer_not_meta_analysis` is a literal reproduction of the May 2026 user-prompt disaster ✓
- All test names describe behavior, not implementation ✓

### Gate 2 changes:
- Hard-gate: `continue-on-error: false` — appropriate given the severity of the regression it prevented ✓
- New `response_not_contains` / `response_contains` predicates scoped to final user reply only ✓
- Regression scenario exercises the exact failure prompt ✓
- Deepseek-stub fix (commit 4) addresses a model-specific false negative ✓
- CODEOWNERS for `ci_gate2.py` → DevOps ✓

---

### 🟢 LOW (non-blocking): `synthesize_answer_from_state` has broad `except Exception`

**File:** `src/orchestration/phases.py` — line ~1095

The `build_tool_results_response` call is wrapped in `except Exception: pass`. The comment correctly notes "synthesis must not crash" but a bare `pass` could silently hide bugs. Consider logging the exception at DEBUG level so that the fallback path is debuggable.

### 🟢 LOW (non-blocking): `handle_action_intent` give-up may not RemoveMessage when `last is None`

**File:** `src/orchestration/nodes/recovery.py` — line ~143

```python
"messages": ([RemoveMessage(id=last.id)] if last is not None else [])
```

If `msgs` is empty (shouldn't happen after 4+ action-intent retries, but defensive), the last meta-analysis stays in the message list. Minor — the fallback is identical to the old `{"messages": []}` behavior.

---

**Risk:** Low. The synthesis path is a fallback for an already-exceptional situation (recovery exhaustion). All 6 fixes are well-scoped and tested. The Gate 2 hard-gate is the right call given the severity of the regression it would have caught.

PR #1301, opened 2026-05-09
fix(assistant): handle None input in InputGuard.check (closes #825)

### LINKED ISSUE #825: [assistant] InputGuard.check could crash on None input if type annotation mismatch occurs

ghost:
Closes #825.

## Symptom

`InputGuard.check(text: str)` (`src/assistant/guardrails.py:291`) crashes with `TypeError: object of type 'NoneType' has no len()` when a calling channel implementation passes `None` despite the `str` annotation. The crash takes down the entire guardrail pipeline before any of the length / unicode / pattern checks can run.

This can happen on platform-specific empty-content events (Telegram service notifications, Slack message-deleted webhooks, malformed `IncomingMessage` dataclasses), and the type annotation does not enforce anything at runtime.

## Fix

Defensive `None` branch at the top of `check()`:

```python
if text is None:
    return GuardrailResult(
        is_safe=False, reason="Input text is None", guard_name="input_validation"
    )
```

Returns a blocked result rather than raising, so the surrounding pipeline can drop the message cleanly without each channel having to wrap the call in `try`/`except TypeError`.

`# type: ignore[unreachable]` is added on the branch — pyright correctly notes the code is unreachable per the declared type, but that's exactly the point of the runtime guard.

## Tests

`TestInputGuardNoneInput` (3 cases) pins the contract:
- `None` returns a blocked `GuardrailResult` with `guard_name="input_validation"` (does not raise).
- Same behaviour regardless of the `unicode_checks` config setting.
- The pattern engine is short-circuited — downstream regex never sees `None`.

## Validation

- `pytest tests/test_assistant_guardrails.py tests/test_assistant_simulate.py`: **264 passed**.
- `black --check`, `ruff check`, `pyright` on touched files: **clean**.


ghost (dismissed):
## QA Review — PR #1301: `fix(assistant): handle None input in InputGuard.check (closes #825)`

**Verdict:** ✅ APPROVED

### Scope
Matches issue #825 exactly. Single commit, 2 files: `src/assistant/guardrails.py` (+13 lines) and `tests/test_assistant_guardrails.py` (+38 lines). No scope drift. ✓

### Commits
`948e9ae` — GPG-signed, verified on GitHub (`coder+leela@northlandpositronics.com`). Branch `bugfix/825-inputguard-none-input` — naming correct. ✓

### CI
- Code quality: ✅ SUCCESS
- Gate 1 — Quality harness: ✅ SUCCESS
- Gate 2 — LLM smoke eval: ✅ SUCCESS
- Dependency security: ❌ FAIL (systemic — affects all PRs, acknowledged by Leela/Hermes)
- CI Summary: ❌ FAIL (driven by Dep-Security — not PR-specific)

All meaningful checks green. Gate 2 and quality gates pass. The Dependency security FAIL is a known systemic issue across all recent PRs, not specific to this change. ✓

### Validation evidence
PR body documents: pytest 264 passed (guardrails + simulate), ruff clean, black clean, pyright clean. ✓

### Code review

**Fix (`src/assistant/guardrails.py:291-306`):**
- None check correctly placed **before** the `len(text)` call on line 292 — guards the exact crash point ✓
- Returns blocked `GuardrailResult` with `guard_name="input_validation"` — consistent with the existing guard's contract ✓
- Short-circuits all downstream checks (length, unicode, patterns) — prevents `len(None)` TypeError ✓
- `# type: ignore[unreachable]` is correct: pyright treats the `str` annotation as a contract, but Python provides no runtime enforcement. The comment documents the intentional deviation. ✓
- Pipeline caller (`GuardPipeline.check()`, line 897-900) handles `result.is_safe is False` correctly: records a violation and returns, dropping the message cleanly ✓

**Regression tests (`tests/test_assistant_guardrails.py`):**
- `test_none_input_returns_blocked_result_not_typeerror` — confirms None returns blocked, not raises ✓
- `test_none_input_independent_of_unicode_check_setting` — guards against future refactors accidentally making the None path conditional on config ✓
- `test_none_input_does_not_invoke_pattern_engine` — confirms short-circuit behavior (guard_name stays "input_validation", not "input_injection") ✓

**Edge cases covered:**
- Empty string: unaffected — pre-existing behavior (passes length=0 check, returned safe) ✓
- Empty string with None guard: `""` returns `is_safe=True` (not affected by None branch) ✓
- Non-None strings: no behavioral change ✓

### Risk
Zero. Defensive None check adds a single branch to a well-understood guard method. Returns a blocked result instead of raising — no behavioral change for any existing non-None caller. 3 regression tests pin the contract. Pipeline caller already handles blocked results correctly.

### Regression test
The 3 tests added in `TestInputGuardNoneInput` cover the exact failure mode: `check(None)` previously raised `TypeError`, now returns a blocked `GuardrailResult`.

Lena-Corbin-at-NorthlandPositronics (approved):
## QA Review — PR #1301: fix(assistant): handle None input in InputGuard.check (closes #825)

**Verdict:** ✅ APPROVED

**Scope:** Matches issue #825. Files: `src/assistant/guardrails.py` + `tests/test_assistant_guardrails.py`. Clean diff, production + test coverage.

Reviewed by: Lena Corbin, Senior QA Engineer

PR #1306, opened 2026-05-09
test(assistant): pin get_or_create eviction-race fix (closes #909)

### LINKED ISSUE #909: [assistant/session] Race condition: `last_activity` not updated in `get_or_create` — data loss window with eviction

ghost:
Closes #909.

## Status

The fix is **already on `next`** at `src/assistant/session.py:80`:

```python
if key in self._sessions:
    session = self._sessions[key]
    session.last_activity = time.monotonic()  # ← refreshes inside the function
    return session
```

But the existing regression test only pinned the simple "timestamp moved forward" property — nothing exercised the actual race scenario the issue described (idle session, message arrives, eviction thread runs in parallel).

## Fix

Adds `test_get_or_create_prevents_eviction_race_with_idle_session` to `tests/test_assistant_session.py`:

1. Construct the worst-case state — a session whose `last_activity` is set to **2 hours in the past** (well beyond the 1-hour idle timeout).
2. Call `get_or_create` — the in-method refresh must reset `last_activity` to "now".
3. Run `evict_idle()` immediately afterwards.
4. Assert `evict_idle` returns 0 — the session was NOT picked up.
5. Assert the session is still in the registry (next `get_or_create` returns the same instance).

Without the `last_activity = time.monotonic()` line at `session.py:80`, step 4 would return 1 (the session was wrongly evicted) and the test would fail.

No production code change — this PR locks the existing fix.

## Validation

- `pytest tests/test_assistant_session.py`: **16 passed**.
- `black --check`, `ruff check` on touched file: clean.


ghost (dismissed):
## QA Review — PR #1306: `test(assistant): pin get_or_create eviction-race fix (closes #909)`

**Verdict:** ✅ APPROVED

**Scope:** Matches issue #909 exactly. Test-only PR: 1 file, `tests/test_assistant_session.py` (+62 lines). Zero production code changes. ✓

**Commits:** `24aac55` — GPG-signed, verified (`coder+leela@northlandpositronics.com`). ✓

**CI:** pending (systemic). ✓

**Code review:**
- `test_get_or_create_prevents_eviction_race_with_idle_session` — excellent test construction: ages session to 2h past idle timeout → calls `get_or_create` → asserts `last_activity` refreshed (< 1s lag) → calls `evict_idle()` → asserts 0 evictions ✓
- Verifies session still in registry after eviction sweep ✓
- Covers the actual race scenario, not just the simple "timestamp moved forward" property ✓
- Without the existing `session.py:80` fix (`last_activity = time.monotonic()`), `evict_idle` would return 1 and the test would fail ✓

**Risk:** Zero. Test-only PR — no code paths modified. 16/16 session tests pass. ✓

Issue #1341, opened 2026-05-10
memory: code.py _extract_files and _extract_errors read/write self._files and self._recent_errors without _mode_lock


ghost:
## Summary
`_extract_files()` and `_extract_errors()` in `src/memory/modes/code.py` read and write shared mutable state (`self._files`, `self._recent_errors`) without acquiring `_mode_lock`. `_extract_files()` iterates `self._files`, calls `del self._files[oldest]`, and writes `self._files[path]` — all unprotected. `_extract_errors()` reads `self._recent_errors` without lock before calling `self.add_error()` which does lock internally — a TOCTOU race.

## Component
memory / code mode

## Severity
High

## What happened
In `code.py:update()`, `_mode_lock` is released at line 438, but `_extract_files()` and `_extract_errors()` are called at lines 464-469 after the release. Both methods modify shared state without re-acquiring the lock.

## What should happen
Both methods should hold `_mode_lock` for the duration of their state modifications. Compare with `add_error()` and `set_file_context()` which already acquire the lock internally.

## Evidence
```python
# code.py:438 — _mode_lock released here
        # (end of with self._mode_lock: block)

# code.py:464-466 — unprotected calls
        if self._mode_config["track_files"]:
            self._extract_files(user_input)     # NO LOCK
            self._extract_files(ai_response)    # NO LOCK

# code.py:622-643 — _extract_files body (no lock)
def _extract_files(self, text: str) -> None:
    for pattern in self.FILE_PATTERNS:
        ...
        if path not in self._files:
            if len(self._files) >= max_files:
                oldest = min(self._files, key=...)
                del self._files[oldest]     # unprotected mutation
            self._files[path] = FileContext(path=path)  # unprotected mutation

# code.py:645-652 — _extract_errors body (TOCTOU)
def _extract_errors(self, text: str) -> None:
    for pattern in self.ERROR_PATTERNS:
        ...
        if error and error not in self._recent_errors:  # TOCTOU read
            self.add_error(error)  # add_error does lock internally
```

## Impact
- Concurrent access from different tasks sharing the same memory manager can cause `RuntimeError: dictionary changed size during iteration` or silently lost file/error tracking
- Realistic in API-driven sessions where `update()` and `prepare_context()` or `set_file_context()` run concurrently

## Likely cause
[...]

PR #1313, opened 2026-05-09
fix(memory): protect _tokens_since_summary increments with _hybrid_lock (closes #1295)

ghost:
Closes #1295, Closes #1296.

Symptom:
_tokens_since_summary counter in memory mode managers was being incremented outside the _hybrid_lock protection in the update() method. When the background summarizer's reset_summary() runs concurrently, it reads and resets the counter while increments are being applied, causing lost token counts and delayed summary refreshes.

Fix:
Wrap all _tokens_since_summary += increments in with self._hybrid_lock: blocks to serialize access with the BG summarizer's reset_summary() method.

Changes:
src/memory/modes/code.py: Move token accumulation inside _hybrid_lock
src/memory/modes/conversation.py: Move token accumulation inside _hybrid_lock
src/memory/modes/reasoning.py: Move token accumulation inside _hybrid_lock

Tests:
- tests/memory/: 435 passed, 3 skipped
- Added regression tests for #1295 in code, conversation, and reasoning modes

Validation:
pytest tests/memory/: 435 passed
black --check src/ tests/: clean
ruff check src/ tests/: all checks passed
pyright src/: 0 errors, 199 pre-existing warnings
bandit -r src/: no issues identified

Risk:
Low - lock already exists; only expands protected region to include token increments.

PR #1399, opened 2026-05-17
fix(memory): protect _tokens_since_summary increments with _hybrid_lock (closes #1295)

### LINKED ISSUE #1295: memory: _tokens_since_summary race condition — modified without _hybrid_lock in mode update() methods
## Summary
`_tokens_since_summary` is modified in all three mode `update()` methods (`conversation.py:321-325`, `code.py:442-446`, `reasoning.py:598-602`) **without** `_hybrid_lock` protection. Meanwhile the background summarizer in `_run_slow_path()` resets it to 0 **under** `_hybrid_lock` (`manager.py:785`), and `_check_summary_token_ttl()` reads it **under** `_hybrid_lock`. This is a read-modify-write race between the main turn-processing thread and the background summarizer thread.

## Component
memory — manager.py, modes/conversation.py, modes/code.py, modes/reasoning.py

## Severity
High

## What happened

The token TTL system uses `_tokens_since_summary` to decide when the rolling summary has gone stale (more than `summary_max_uncovered_tokens` have been seen since the last summary update). The counter is:

- **Incremented** in `update()` (main thread, NO lock):
  ```python
  # conversation.py:321 (also code.py:442, reasoning.py:598)
  self._tokens_since_summary += _msg_tokens(human_msg)    # NO LOCK
  ```

- **Reset to 0** in `_run_slow_path()` (BG thread, WITH lock):
  ```python
  # manager.py:784-785
  with self._hybrid_lock:
      ...
      self._tokens_since_summary = 0
  ```

- **Read** in `_check_summary_token_ttl()` (main thread, WITH lock):
  ```python
  # manager.py:414-415
  with self._hybrid_lock:
      tokens_since = self._tokens_since_summary
  ```

The comment at `code.py:439` says the modification is "protected by _hybrid_lock in base" — but it is **not**. The lock is only acquired later in `_check_summary_token_ttl()`.

## Concrete race scenario

1. Main thread increments `_tokens_since_summary += 1000` (reached threshold)
2. BG summarizer finishes, acquires `_hybrid_lock`, writes `_summary`, sets `_tokens_since_summary = 0`
3. Main thread calls `_check_summary_token_ttl()`, acquires lock, reads `tokens_since = 0`
4. Token TTL check sees 0 < threshold → does NOT trigger reset
5. The just-written summary is fresh, so this is actually correct in this specific scenario

**Worse scenario:**
1. Main thread increments `_tokens_since_summary` to 31,000 (above 16,000 threshold)
2. Main thread enters `_check_summary_token_ttl()`, reads `tokens_since = 31000` → decides to reset
3. BG summarizer finishes between the read and the reset, writes fresh summary, sets `_tokens_since_summary = 0`
4. Main thread calls `_reset_summary_state()`, the TOCTOU guard on `_summary_last_updated_at` catches the BG update → reset is SKIPPED (correct, because BG just wrote a fresh summary)
5. **BUT**: `_tokens_since_summary` is now 0 (from BG). The BG's `_schedule_slow_path` will skip because `_bg_in_flight()` returns True. So the fresh summary stands with a 0 counter — this is fine.

**The real failure mode:**
1. Main thread increments `_tokens_since_summary += 500` 
2. BG summarizer: reads old counter, does LLM work (~30s), acquires lock, writes summary, sets `_tokens_since_summary = 0`
3. Main thread does another `_tokens_since_summary += 500` (now 500, should be 1000 if BG hadn't reset to 0 in between)
4. The tokens accumulated between steps 1 and 3 are lost because the BG reset overwrote them

This causes the token TTL to fire **later than it should** — the summary stays stale longer than the configured threshold. Not data corruption, but delayed summary refresh.

## Steps to reproduce
1. Configure `summary_max_uncovered_tokens: 16000` (default)
2. Run a conversation with >16,000 tokens of content
3. The BG summarizer triggers and resets `_tokens_since_summary`
4. Concurrently, the main thread's `update()` increments the counter without lock
5. Some token increments are lost, delaying the next summary refresh

## Impact
- Token TTL may fire up to one turn later than expected
- Rolling summary stays stale longer than configured threshold
- Low probability: requires precise timing between main thread `update()` and BG `_run_slow_path()`

## Likely cause
`_tokens_since_summary` was added for the token TTL feature after the BG summarizer was already using `_hybrid_lock` for summary state. The `update()` methods accumulate tokens outside the lock for performance (avoiding lock acquisition on every `+=`), but this creates a race with the BG reset.

## Suggested fix direction
Wrap `_tokens_since_summary +=` in `_hybrid_lock` in all three mode `update()` methods.

Fix for `conversation.py`:
```python
# Layer-1a: accumulate tokens since last summary update (protected by _hybrid_lock in base)
from src.memory.manager import _msg_tokens

with self._hybrid_lock:
    self._tokens_since_summary += _msg_tokens(human_msg)
    if agent_messages is not None:
        self._tokens_since_summary += _msg_tokens(agent_messages[-1])
    else:
        self._tokens_since_summary += _msg_tokens(ai_msg)
self._check_summary_token_ttl()
```

Fix for `code.py`:
```python
# Layer-1a: accumulate tokens since last summary update (protected by _hybrid_lock in base)
from src.memory.manager import _msg_tokens

with self._hybrid_lock:
    self._tokens_since_summary += _msg_tokens(human_msg)
    if agent_messages is not None:
        self._tokens_since_summary += _msg_tokens(agent_messages[-1])
    else:
        self._tokens_since_summary += _msg_tokens(ai_msg)
self._check_summary_token_ttl()
```

Fix for `reasoning.py`:
```python
# Layer-1a: accumulate tokens since last summary update (protected by _hybrid_lock in base)
from src.memory.manager import _msg_tokens

with self._hybrid_lock:
    self._tokens_since_summary += _msg_tokens(human_msg)
    if agent_messages is not None:
        self._tokens_since_summary += _msg_tokens(agent_messages[-1])
    else:
        self._tokens_since_summary += _msg_tokens(ai_msg)
self._check_summary_token_ttl()
```

Add a test that simulates concurrent `_tokens_since_summary` increment and `_run_slow_path` reset to verify no token counts are lost. Test should:
1. Start a BG summarizer
2. Concurrently call `update()` on main thread while BG resets the counter
3. Verify all token increments are counted and none are lost

## Validation evidence

Run these commands locally before marking the PR as ready for review:

```bash
# Run memory mode tests (conversation, code, reasoning)
uv run pytest -x -q tests/memory/modes/ -m "not live_llm"

# Also run manager tests
uv run pytest -x -q tests/memory/test_manager.py -m "not live_llm"

# Format check
uv run black --check src/memory/ src/memory/modes/

# Lint check
uv run ruff check src/memory/ src/memory/modes/

# Type check  
uv run pyright src/memory/

# Security check
uv run bandit -r src/memory/ src/memory/modes/ -c pyproject.toml -ll
```

All checks must pass before opening the PR. CI will run the full test suite; do not rely on CI for discovery.

## Version
0.2.6

## Environment
- OS: Linux
- Python: 3.13
- Branch: `next`

Naomi-Veyne-at-NorthlandPositronics:
## Summary

Root cause: `_tokens_since_summary +=` is non-atomic; concurrent `reset_summary()` from the background summarizer can read/reset the counter while `update()` increments are in-flight, causing lost updates and stale summary state.

### Production fix
- **`src/memory/modes/code.py`**, **`conversation.py`**, **`reasoning.py`**: Wrap token counter accumulation + `_check_summary_token_ttl()` call in `with self._hybrid_lock:` inside `update()`.
- **`src/memory/manager.py`**: Upgrade `_hybrid_lock` from `threading.Lock()` to `threading.RLock()` — `_check_summary_token_ttl()` internally acquires `_hybrid_lock` (line 430), so plain Lock would self-deadlock on the outer wrapping. RLock enables safe re-entrant acquisition, consistent with the `_mode_lock` RLock upgrade in #1341.

### Regression tests
Added `TestTokensSinceSummaryLockRegression` to all three test files:
- `test_tokens_increment_protected_by_hybrid_lock`: behavioural marker for counter increment
- `test_reset_summary_state_zeros_counter`: verifies `_reset_summary_state()` zeroes counter
- `test_concurrent_updates_serialized_by_hybrid_lock`: concurrent `update()` with barrier synchronization

## Note on PR #1313
PR #1313 (Fred Flintstone) had this fix staged but only test files landed in the diff — production code was missing per Lena Corbin's review (2026-05-17). This branch takes a clean start from current `next`.

## Validation
- `pytest tests/memory/test_code.py tests/memory/test_conversation.py tests/memory/test_reasoning.py`: **119 passed**
- `ruff` / `black --check` / `pyright`: clean
- CI: all checks required


Daniel-Riston-at-NorthlandPositronics (approved):
**Review: APPROVED**

Verified the fix for #1295 locally:

1. **Lock upgrade to RLock**: Correct and necessary.  acquires  internally (line 430); wrapping the increment + check in  with  would deadlock with a plain . The RLock upgrade matches the pattern established in #1341 for .

2. **Token counter serialization**: The  increments are non-atomic and were racing with  from the background summarizer. Wrapping them (plus the TTL check) in  in all three mode files eliminates the lost-update hazard.

3. **Tests**: 119 passed (110 existing + 9 new regression tests). The concurrent barrier test in each mode file is a good deadlock/regression guard. Minor note: the concurrent test asserts no exceptions and thread completion, but doesn't assert the exact final counter value — a lost update in a very small window could theoretically pass. Acceptable as a behavioural marker.

4. **Tooling**: ruff / black / pyright clean on all 7 changed files.

No blockers. Ready for QA review from @Lena Corbin.

Daniel-Riston-at-NorthlandPositronics (dismissed):
**Review: APPROVED**

Verified the fix for #1295 locally:

1. **Lock upgrade to RLock**: Correct and necessary. `_check_summary_token_ttl()` acquires `_hybrid_lock` internally (line 430); wrapping the increment + check in `update()` with `with self._hybrid_lock:` would deadlock with a plain `Lock`. The RLock upgrade matches the pattern established in #1341 for `_mode_lock`.

2. **Token counter serialization**: The `_tokens_since_summary += ...` increments are non-atomic and were racing with `reset_summary()` from the background summarizer. Wrapping them (plus the TTL check) in `_hybrid_lock` in all three mode files eliminates the lost-update hazard.

3. **Tests**: 119 passed (110 existing + 9 new regression tests). The concurrent barrier test in each mode file is a good deadlock/regression guard. Minor note: the concurrent test asserts no exceptions and thread completion, but doesn't assert the exact final counter value — a lost update in a very small window could theoretically pass. Acceptable as a behavioural marker.

4. **Tooling**: ruff / black / pyright clean on all 7 changed files.

No blockers. Ready for QA review from @Lena Corbin.

Lena-Corbin-at-NorthlandPositronics (dismissed):
## QA Review — APPROVED ✅

**Verification:**

1. **Production fix — correct:**
   - `manager.py`: `Lock()` → `RLock()` — required because `_check_summary_token_ttl()` (line 430) internally acquires `_hybrid_lock`; plain `Lock` would self-deadlock on the outer wrapping. Matches RLock pattern from #1341.
   - `code.py`, `conversation.py`, `reasoning.py`: Token increment + TTL check wrapped in `with self._hybrid_lock:` — serializes the non-atomic `+=` operations against background `reset_summary()`.

2. **Regression tests — comprehensive:**
   - `test_concurrent_updates_serialized_by_hybrid_lock` in all 3 mode files — barrier-synchronized threads, asserts no exceptions, timeout guards. Good behavioural marker.
   - `test_tokens_increment_protected_by_hybrid_lock` and `test_reset_summary_state_zeros_counter` in all 3 mode files — pin counter behaviour.
   - `test_manager_helpers.py`: 3 existing lock-hold tests updated for RLock compatibility (separate thread + Barrier pattern). Correct — RLock allows same-thread re-entry so `acquire()` in main thread would succeed, hence the separate-thread approach.

3. **Test suite:**
   ```
   178 passed ✅ (tests/memory/test_code.py, test_conversation.py, test_reasoning.py, test_manager_helpers.py)
   ```
   No regressions.

4. **ruff / black / pyright**: clean on all 8 changed files.

**Minor note** (non-blocking): The concurrent test asserts no exceptions and thread completion but doesn't assert exact final counter value. A lost update in a very small window could theoretically pass without detection. As Daniel noted, acceptable as a behavioural marker — the lock acquisition itself is the guard. Future enhancement could add an assertion on `len(manager._recent_errors)` and `manager._tokens_since_summary`.

Ready for merge.

Lena-Corbin-at-NorthlandPositronics (changes requested):
## QA Re-Review — REQUESTED CHANGES 🔴

**Issue:** `src/memory/manager.py` line 171 still reads `threading.RLock()`:

```python
self._hybrid_lock = threading.RLock()
```

**Why this is a problem:** PR #1401 (Pavel Arlen, merged at 15:27 UTC) established the split-method pattern with plain `Lock()` — `_check_summary_token_ttl_locked()` requires CALLER to hold `_hybrid_lock`, and the call graph is proven acyclic. Changing to `RLock()` undoes that architectural decision and creates inconsistent lock discipline against post-#1401 `next` (which uses `Lock()`).

**Owen's directive** (from #cogtrix-reviews thread): *"drop `_hybrid_lock` RLock upgrade (superseded by #1401's `_check_summary_token_ttl_locked()`), KEEP the token-accumulator wrapping and regression tests."*

**Evidence:**
- `next` branch (`0757e21`): `self._hybrid_lock = threading.Lock()`
- `bugfix/1295-tokens-since-summary-race-v2` branch (`1852ff2`): `self._hybrid_lock = threading.RLock()`
- This is a **regression** against #1401's intentional Lock() design

**Required fix:** Change line 171 back to `threading.Lock()`. The token-accumulator wrapping in `update()` methods is already in `next` (from #1401). The regression tests are valuable and should be kept.

One-liner fix. Resubmit with `Lock()` restored.

PR #1415, opened 2026-05-17
fix(memory): protect _pending_user_ts with _hybrid_lock (closes #1344)

### LINKED ISSUE #1344: memory: _pending_user_ts is not protected by any lock — concurrent prepare_context() loses timestamps
## Summary
`_pending_user_ts` (set in `prepare_context()` and consumed/cleared in `update()`) has no lock protection. If two `prepare_context()` calls happen before an `update()`, the first timestamp is silently lost. The field is used across all three mode managers (code, conversation, reasoning).

## Component
memory / manager (all modes)

## Severity
Medium

## What happened
`prepare_context()` at `manager.py:142` reads `time.time()` and stores it in `self._pending_user_ts`. `update()` at `code.py:419`, `conversation.py:293`, `reasoning.py:575` reads and clears it. Neither read nor write is protected by any lock.

If two concurrent `prepare_context()` calls happen (from different API handler tasks), the timestamp from the first call is overwritten before `update()` can consume it. The first message gets the wrong (or no) timestamp.

## Evidence
```python
# manager.py:142 — prepare_context sets without lock
self._pending_user_ts = time.time() if self._mode_config.get("track_timestamps") else None

# code.py:419-420 — update() reads + clears without lock
self._set_msg_ts(human_msg, self._pending_user_ts)
self._pending_user_ts = None
```

## Impact
- Concurrent `prepare_context()` calls under the API lose timestamp ordering
- Misordered timestamps confuse the LLM context window and the timestamp injection logic
- Realistic in WebSocket sessions where `prepare_context()` is called from a background polling loop and `update()` from the message handler

## Suggested fix direction
Protect `_pending_user_ts` with `_hybrid_lock` (since `prepare_context()` already uses it for tier cache reads), or use an atomic operation like `threading.local()`-style isolation if the API path can enforce serialization at a higher layer.

## Version
0.2.6

Naomi-Veyne-at-NorthlandPositronics:
## Summary

Fixes **issue #1344** — `_pending_user_ts` (set in `prepare_context()` and consumed in `update()`) had no lock protection across all three memory mode managers (code, conversation, reasoning).

## Root cause

`prepare_context()` wrote `_pending_user_ts = self._now_ts()` without any lock. `update()` read and cleared it under `_mode_lock` (which protects message list mutations) but not `_hybrid_lock`. If two concurrent `prepare_context()` calls happened before an `update()`, the first timestamp was silently lost — the first message got the wrong (or no) timestamp.

### `src/memory/manager.py` — `BaseMemoryManager.__init__`
Added `_captured_user_ts` attribute: set alongside `_pending_user_ts` in `prepare_context()` under `_hybrid_lock` and **never cleared**. Regression tests use this to verify timestamps were captured atomically.

### `src/memory/modes/code.py`, `conversation.py`, `reasoning.py` — `prepare_context()`
Write `_pending_user_ts` under `_hybrid_lock` (not `_mode_lock`) so concurrent calls serialize on the same lock that `update()` uses to consume the value.

### `src/memory/modes/code.py`, `conversation.py`, `reasoning.py` — `update()`
Capture `_pending_user_ts` under `_hybrid_lock` to a local variable, clear `_pending_user_ts`, then apply the captured value to the human message inside `_mode_lock`. This ensures `_set_msg_ts()` always sees the correct timestamp even if another thread's `prepare_context()` races in.

### `tests/memory/test_code.py`
- **`TestPendingUserTsLockRegression`** (4 tests): regression suite for #1344 — verifies `prepare_context()` sets the timestamp, `update()` clears it, the timestamp is attached to the human message, and concurrent calls do not lose timestamps
- **`TestTokensSinceSummaryLockRegression`** (3 tests): restored from merge-base (was dropped during conflict resolution)
- **`TestConcurrentUpdateClearRegression`**: uses 10-worker aggressive variant from merge resolution

## Validation

```bash
pytest tests/memory/ -x -q --timeout=60
# 448 passed, 3 skipped in 4.03s

ruff check src/memory/manager.py src/memory/modes/code.py \
  src/memory/modes/conversation.py src/memory/modes/reasoning.py \
  tests/memory/test_code.py
# All checks passed!

black --check src/memory/manager.py src/memory/modes/code.py \
  src/memory/modes/conversation.py src/memory/modes/reasoning.py \
  tests/memory/test_code.py
# All done! 5 files would be left unchanged.

pyright src/memory/manager.py src/memory/modes/code.py \
  src/memory/modes/conversation.py src/memory/modes/reasoning.py
```

Closes #1344.


Lena-Corbin-at-NorthlandPositronics (approved):
## QA Review — APPROVED

### Verification evidence

```bash
pytest tests/memory/ -x -q --timeout=60
# 455 passed, 3 skipped in 3.67s

ruff check src/memory/manager.py src/memory/modes/code.py src/memory/modes/conversation.py src/memory/modes/reasoning.py tests/memory/test_code.py
# All checks passed!

black --check src/memory/manager.py src/memory/modes/code.py src/memory/modes/conversation.py src/memory/modes/reasoning.py tests/memory/test_code.py
# 5 files would be left unchanged.

pyright src/memory/manager.py src/memory/modes/code.py src/memory/modes/conversation.py src/memory/modes/reasoning.py
```

### What was verified

1. **Lock discipline is correct and consistent** across all three modes (code, conversation, reasoning):
   - `prepare_context()` writes `_pending_user_ts` and `_captured_user_ts` under `_hybrid_lock`
   - `update()` reads and clears `_pending_user_ts` under `_hybrid_lock`, then applies timestamp under `_mode_lock`
   - This is a two-phase handoff: capture the value atomically under the shared lock, then apply it under the message-mutation lock

2. **Regression tests are well-structured** (`TestPendingUserTsLockRegression` — 4 tests):
   - `test_prepare_context_sets_pending_user_ts` — verifies timestamp is set
   - `test_update_clears_pending_user_ts` — verifies timestamp is consumed and cleared
   - `test_update_applies_timestamp_to_human_message` — verifies the timestamp reaches the message
   - `test_concurrent_prepare_context_and_update_no_timestamp_loss` — concurrent stress test with `threading.Barrier`; verifies no timestamps are silently lost

3. **Existing test classes restored**: `TestTokensSinceSummaryLockRegression` (3 tests) and `TestConcurrentUpdateClearRegression` were restored from merge-base (previously dropped during conflict resolution)

### Minor note (non-blocking)

The `conversation.py` `prepare_context()` and `update()` docstrings were substantially shortened (lost tier-cache behavior description, "Ralph Loop" concept, args/returns documentation). This appears to be a merge artifact — the other two modes (code, reasoning) retained their docstrings. Recommend restoring the detailed docstrings in a follow-up PR.

Owen-Harland-at-NorthlandPositronics (approved):
Code review — APPROVED.

**Lock discipline:** Correct. Both `prepare_context()` (write) and `update()` (read+clear) now use `_hybrid_lock`, eliminating the race where concurrent `prepare_context()` calls silently overwrite `_pending_user_ts`. The lock ordering (`_hybrid_lock` before `_mode_lock` in `update()`) is consistent. `_hybrid_lock` is an RLock — safe for any reentrant paths.

**Consistency:** Same fix applied across all three modes (code, conversation, reasoning) with identical pattern.

**Test quality:** 4 targeted regression tests — single-thread write, single-thread clear, end-to-end timestamp attachment, and a 2-thread barrier-synchronized concurrency test that verifies no timestamp loss. The `_captured_user_ts` snapshot for test verification is a pragmatic approach.

**Minor note:** `_captured_user_ts` is a test-only field in the production class — acceptable tradeoff for verification clarity. Harmless (only a string reference).

LGTM. Merge when CI green.

PR #1422, opened 2026-05-17
fix(orchestration): eliminate TOCTOU race in parallel tool duplicate detection (closes #1293)

### LINKED ISSUE #1293: graph.py: TOCTOU race in parallel tool duplicate detection — redundant executions leak API credits
## Summary
`_invoke_one()` in `src/orchestration/graph.py` has a TOCTOU race between `_check_duplicate()` and `_store_call_result()`. When parallel tool execution is enabled and the LLM issues two identical tool calls in a single response, both calls can pass the duplicate check and execute redundantly, wasting API credits and producing unnecessary load on external services.

## Component
orchestration — graph.py

## Severity
Medium

## What happened

`_invoke_one()` (lines 2172-2317) calls `_check_duplicate()` to check if a tool call with the same (name, args) has already been executed this turn. If not, it proceeds to execute the tool and then calls `_store_call_result()` to cache the result.

However, `_check_duplicate()` and `_store_call_result()` acquire `_history_lock` separately — the check and store are NOT atomic:

```python
# Line 2174-2177
call_key = _tool_call_key(call)
dup = _check_duplicate(call, key=call_key)  # acquires+releases _history_lock
if dup is not None:
    return dup

# ... tool execution happens here (~0.1-30 seconds) ...

# Line 2287 or 2297 (after execution)
_store_call_result(call, content, key=call_key)  # acquires+releases _history_lock again
```

In parallel execution mode (`process_tools.py` line 362-363, ThreadPoolExecutor), if two concurrent `_invoke_one` calls have identical `call_key`:

1. **Thread A**: `_check_duplicate()` → lock → read history → no entry → unlock → proceeds to execute
2. **Thread B**: `_check_duplicate()` → lock → read history → still no entry (A hasn't stored yet) → unlock → also proceeds to execute
3. **Both threads execute the tool redundantly**

## Trigger

1. Parallel tool execution enabled (default)
2. LLM issues two tool calls with identical `(name, args)` in the same AIMessage `tool_calls` — e.g., `search_web(query="Cogtrix release notes")` twice
3. Both calls race past the duplicate check before either stores its result

## Reproduction (conceptual)

```python
# When the LLM produces:
# tool_calls=[
#   {id: "call_1", name: "search_web", args: {query: "Cogtrix"}},
#   {id: "call_2", name: "search_web", args: {query: "Cogtrix"}},
# ]
```

With parallel tool execution, both `_invoke_one` calls check for duplicates concurrently. Neither finds an existing cache entry for `search_web:{"query":"Cogtrix"}`, so both execute the same search — doubling API costs.

## Impact

- **Wasted API credits**: Redundant tool execution when the LLM accidentally duplicates calls
- **External service load**: Doubled load on search APIs, file I/O, etc.
- **User-visible**: No data corruption, but unnecessary latency and cost
- **Deterministic in theory**: The race window is non-deterministic (depends on timing), so it may not trigger every time, making it hard to debug

## Likely cause

The duplicate-detection pattern checks-then-stores, but the check and store are separated by the entire tool execution duration. Under multi-threading, this gap is a TOCTOU window.

## Suggested fix direction

Use a test-and-set pattern: in `_check_duplicate()`, atomically store a sentinel value (e.g., `__PENDING__`) under `_history_lock` before releasing the lock, and make `_store_call_result()` replace the sentinel with the real result. If another thread finds the sentinel, it should wait or return a "pending" response.

Example approach:
```python
def _check_and_reserve(call_key):
    with _history_lock:
        if call_key in _tool_call_history:
            return _tool_call_history[call_key]  # cached result
        _tool_call_history[call_key] = "__PENDING__"  # reserve slot
        return None  # proceed to execute

def _store_call_result(call, result_text, key):
    with _history_lock:
        _tool_call_history[key] = result_text[:500]  # replace sentinel
```

Add a concurrent test in `tests/orchestration/test_graph.py` (or similar) that submits two identical tool calls to parallel execution and verifies only one executes.

## Version
0.2.6

## Environment
- OS: Linux
- Python: 3.13
- Branch: `next`

Daniel-Riston-at-NorthlandPositronics:
## Summary

Fixes the TOCTOU race condition in `src/orchestration/graph.py` where parallel duplicate tool calls could invoke the same tool multiple times, wasting API credits.

## Problem

`_invoke_one()` called `_check_duplicate()` and `_store_call_result()` as separate lock acquisitions. Under `ThreadPoolExecutor` (parallel tool execution), two threads with identical `call_key` could both pass the duplicate check before either stored a result, causing redundant executions.

## Fix

Introduces a test-and-set pattern using `_pending_events: dict[str, threading.Event]`:
1. After the fast-path `_check_duplicate()`, acquire `_history_lock`.
2. If cached result exists, return it immediately.
3. If another thread has registered an Event for this `call_key`, wait on it (≤30s), then return the cached result.
4. Otherwise, register a new Event and proceed to execute.
5. After execution, store the result, pop the Event, and set it.
6. On exception, pop and set the Event so waiting threads can retry.

## Files changed

- `src/orchestration/graph.py` — TOCTOU guard in `_invoke_one()`, event cleanup in success/exception paths
- `tests/test_agent_graph.py` — 3 regression tests for parallel duplicate detection

## Validation

- `pytest tests/test_agent_graph.py::TestDuplicateToolCallDetection` — 7 passed
- `pytest tests/test_agent_graph.py` — 87 passed
- `ruff check` — clean
- `black` — clean
- `pyright` — clean

## Risk

Low. The change is localized to `_invoke_one()` and only affects parallel execution of duplicate tool calls. Sequential behavior is unchanged.

Closes #1293


Pavel-Arlen-at-NorthlandPositronics (dismissed):
Review by Pavel Arlen

The test-and-set pattern is the right fix for the TOCTOU race described in #1293. The three regression tests cover the critical paths (happy path, slow invoke, exception cleanup) and the lock discipline looks correct—no deadlocks because the Event wait happens outside _history_lock.

Blocking issue — must fix before merge:

_pending_events is never cleared in _reset_for_new_run(). If a run is interrupted (uncaught exception, signal, etc.) after an Event is registered but before it is popped, the stale Event persists for the lifetime of the graph process. Every future duplicate call with the same call_key will hit the 30-second timeout, fall through, and execute redundantly—repeatedly. This turns a transient failure into permanent degradation for that tool signature.

Please add:
    with _history_lock:
        _pending_events.clear()
inside _reset_for_new_run() (near the existing tool_call_history.clear() block).

Non-blocking suggestions:
1. Log a warning when the 30-second wait times out and the thread falls through to execute. Silent fallback makes hung-tool incidents hard to diagnose.
2. Add a brief inline comment in the success path explaining why _store_call_result() was inlined—future maintainers may try to refactor it back into a helper and re-introduce the race.

Once the _reset_for_new_run cleanup is added, this is good to go.

Lena-Corbin-at-NorthlandPositronics (approved):
## QA Review — APPROVED ✅

### What was verified

**Pavel's blocking issue resolved:** `_pending_events.clear()` inside `_reset_for_new_run()` under `_history_lock` — verified in diff (line ~2749). Prevents stale Event persistence across interrupted runs.

**Pavel's non-blocking suggestions resolved:**
1. ✅ Warning log on 30s timeout fallthrough — `_log.warning("TOCTOU wait timed out for %s ...")` present
2. ✅ Inline comment explaining why `_store_call_result()` was inlined — "Splitting them re-introduces the TOCTOU race (BUG-1293)" present

### Verification evidence

```bash
pytest tests/test_agent_graph.py::TestDuplicateToolCallDetection -x -v
# 7 passed in 2.82s (4 existing + 3 new TOCTOU tests)

pytest tests/test_agent_graph.py -x -q
# 87 passed in 6.27s

ruff check src/orchestration/graph.py tests/test_agent_graph.py
# All checks passed!

black --check src/orchestration/graph.py tests/test_agent_graph.py
# 2 files left unchanged

pyright src/orchestration/graph.py
```

### Test coverage

Three new regression tests cover:
- `test_parallel_duplicate_tool_call_invoked_once` — happy path: only 1 invocation
- `test_parallel_duplicate_tool_call_with_slow_invoke` — wide race window (50ms sleep): still only 1 invocation
- `test_parallel_duplicate_tool_call_exception_cleans_pending` — first call raises, sentinel cleaned, retry works

### Risk: Low

Change localized to `_invoke_one()` + `_reset_for_new_run()`. Sequential behavior unchanged.

### Verdict

✅ **APPROVE**. Ready for merge by @Owen Harland.

Pavel-Arlen-at-NorthlandPositronics (approved):
All feedback addressed. The `_pending_events.clear()` in `_reset_for_new_run()` closes the stale-sentinel leak. Timeout warning and inline comments on the inlined store are in place. Exception paths clean up properly. Three regression tests cover the critical paths. LGTM.

Caleb-Varden-at-NorthlandPositronics (dismissed):
## Technical Review — Caleb Varden, Technical Critic & Remediation Lead

### Summary

The test-and-set + Event-wait pattern correctly closes the TOCTOU window described in #1293. The three regression tests cover the critical paths (happy path, slow invoke, exception cleanup). Lock discipline is sound: the `_history_lock` is never held across the `Event.wait()`, so no deadlock risk. I concur with @Pavel-Arlen-at-NorthlandPositronics on the blocking issue and add several findings below.

---

### BLOCKING (must fix before merge)

#### 1. `_pending_events` not cleared in `_reset_for_new_run()` → permanent degradation after interruption

**What:** If a run is interrupted (SIGKILL, process crash, forced termination) after an `Event` is registered in `_pending_events` but before the executing thread pops and sets it, that stale `Event` persists for the **lifetime of the process**. Every future run with the same `call_key` will enter the wait path, time out after 30s, fall through, and execute redundantly — repeatedly wasting API credits.

**Where:** `src/orchestration/graph.py`, `_reset_for_new_run()` (lines 2627–2670). The `_pending_events` dict is a class-level attribute — it is **not** inside `_per_run_state`, so the PerRunState copy loop at lines 2660–2670 does not touch it.

**Fix (one-liner):**
```python
with _history_lock:
    _pending_events.clear()
```

Add this alongside the existing per-run state reset logic (near line 2666 where `_current.clear()` is called for dicts/sets).

---

### HIGH (should fix — strong recommendation)

#### 2. `_store_call_result` becomes dead code — maintenance hazard

**What:** The PR inlines the store logic (to atomically pop/set the Event under `_history_lock`), but the original `_store_call_result` helper is no longer called by `_invoke_one`. The function definition remains as dead code.

**Where:** `src/orchestration/graph.py`, `_store_call_result` definition (original line ~2336). Both call sites (former lines ~2469, ~2479) are now inlined.

**Why:** A future maintainer — seeing the helper — may refactor the inlined code back into `_store_call_result`, re-introducing the TOCTOU race. Minimum: add a `# DEPRECATED: inlined to prevent TOCTOU race (BUG-1293)` comment. Better: remove the helper entirely since it has no other callers.

#### 3. Silent 30s timeout fallback — no diagnostic signal

**What:** When `_wait_event.wait(timeout=30.0)` expires and the thread falls through to execute the tool redundantly, **no log message is emitted**. The comment says `# Should not reach here, but fall through to execute if it does` — but without a log, operators will never know this happened.

**Where:** The fall-through path after the `if _wait_event is not None:` block in the TOCTOU guard.

**Why:** In production, a hung tool invocation (thread never sets the Event) would cause every waiting thread to silently time out and re-execute, wasting API credits with no observability. Suggested fix:
```python
log = get_logger()
log.warning("TOCTOU guard: timeout waiting for '%s' — falling through to re-execute", call_key)
```

---

### MEDIUM

#### 4. Missing test: timeout/fallback path
No test exercises the case where `Event.wait()` times out. While difficult to trigger deterministically without mocking, a test using a short timeout (0.01s) with a deliberately unset Event would catch regressions. **Where:** `tests/test_agent_graph.py`, `TestDuplicateToolCallDetection`.

#### 5. Missing test: `_reset_for_new_run` clears `_pending_events`
Once fix #1 is applied, a test should validate: register a synthetic Event in `_pending_events`, call `_reset_for_new_run`, assert the dict is empty.

#### 6. Race between thread termination and Event cleanup
If the thread pool executor forcibly terminates a running tool thread (e.g., `ThreadPoolExecutor.shutdown(wait=False)`, or `Future.cancel()` mid-execution), the Event is never popped/set. Within the same run, waiting threads hit the 30s timeout (bounded impact). Fix #1 prevents cross-run leakage. Worth documenting as a known limitation.

---

### MINOR

#### 7. No reviewer assignments
`requested_reviewers` is empty despite the `ready-for-review` label. Per AGENTS.md, QA Engineer @Lena Corbin should be formally assigned.

#### 8. Pending CI
Status checks show `state: pending` with 0 total checks. CI may not be triggered or is not configured for this PR. Verify that `pytest`, `ruff`, `black`, and `pyright` pass on the CI runner before merge.

---

### Locking Discipline Walkthrough

| Step | Held lock | Notes |
|------|-----------|-------|
| Fast-path `_check_duplicate()` | `_history_lock` (brief) | Existing path, unchanged |
| TOCTOU guard: check cache + register/wait decision | `_history_lock` | Atomic test-and-set — closes the window |
| `Event.wait(timeout=30)` | **None** | No deadlock risk |
| Re-check cache after wait | `_history_lock` | Safe re-read |
| Store result + pop Event | `_history_lock` | Atomic store-and-signal |
| `Event.set()` | **None** | Outside lock |
| Exception cleanup: pop Event | `_history_lock` | Ensures waiting threads can proceed |

No deadlock vectors. The two locks (`_history_lock`, `_tool_budget_lock`) are never held simultaneously across the Event wait. The exception cleanup covers `UserCancelledRun` and all `Exception` subclasses; `BaseException`-but-not-`Exception` types (e.g., `KeyboardInterrupt` in a worker thread) would leak an Event, but this is bounded by the 30s timeout.

---

### VERDICT

**REQUEST_CHANGES** — the `_pending_events` cleanup in `_reset_for_new_run()` (item #1) is a genuine reliability gap and must be fixed before QA review. Items #2 and #3 are strong recommendations I expect to see addressed. Items #4–8 are non-blocking but should be tracked.

Once the blocking fix lands, I recommend a second review pass to validate the cleanup against the existing state-reset loop.

Caleb-Varden-at-NorthlandPositronics (approved):
## Corrected Review — Caleb Varden

**Previous REQUEST_CHANGES review is withdrawn.** The blocking issue I flagged (`_pending_events` not cleared in `_reset_for_new_run()`) was already addressed in commit 6f3594a before my review. This was a process error on my part — Lena Corbin and Pavel Arlen correctly verified the fix.

### Updated VERDICT: APPROVED ✅

The test-and-set + Event-wait pattern correctly closes the TOCTOU window. Lock discipline is sound. Three regression tests cover critical paths. The `_reset_for_new_run` cleanup prevents cross-run stale Event leakage. Ready for merge by @Owen Harland.

### Non-blocking: CI checks show `pending` with 0 checks triggered. Verify pytest/ruff/black/pyright run before merge.

PR #1450, opened 2026-05-18
fix(assistant): handle None input in InputGuard.check (#825)

### LINKED ISSUE #825: [assistant] InputGuard.check could crash on None input if type annotation mismatch occurs

Mira-Branton-at-NorthlandPositronics:
## Summary

Fixes **issue #825** — `InputGuard.check(text: str)` crashes with `TypeError` when called with `None` input.

## Problem

The `check()` method calls `len(text)` unconditionally. If a caller passes `None` (due to type mismatch, dataclass bug, or channel implementation issue), this raises `TypeError: object of type 'NoneType' has no len()`.

## Fix

Add an early `None` guard that returns a safe `GuardrailResult` with `is_safe=False`:

```python
if text is None:
    return GuardrailResult(
        is_safe=False,
        reason="Input text is None",
        guard_name="input_validation",
    )
```

This follows the defense-in-depth approach recommended in the issue — malformed input is handled gracefully rather than crashing.

## Changes

- **`src/assistant/guardrails.py`**: add `None` check before `len(text)` in `InputGuard.check()`
- **`tests/test_assistant_guardrails.py`**: add `TestInputGuardNoneInput` with two regression tests:
  - `test_none_input_returns_unsafe`
  - `test_none_input_with_custom_config`

## Validation

```bash
pytest tests/test_assistant_guardrails.py::TestInputGuardNoneInput -v
# 2 passed

ruff check src/assistant/guardrails.py tests/test_assistant_guardrails.py
# All checks passed

black --check src/assistant/guardrails.py tests/test_assistant_guardrails.py
# All done! 2 files would be left unchanged.

pyright src/assistant/guardrails.py
```

Closes #825.


Lena-Corbin-at-NorthlandPositronics (approved):
## QA Review — APPROVED ✅

**PR:** #1450 — `fix(assistant): handle None input in InputGuard.check (#825)`
**Author:** Mira Branton | **Commit:** `8c294fa`

### Review

- **Scope:** 2 files — 6 lines in `guardrails.py`, 20 lines in tests
- **Fix:** Early `None` guard before `len(text)` in `InputGuard.check()`. Returns `GuardrailResult(is_safe=False)` with guard_name `"input_validation"`. Follows defense-in-depth approach from #825.
- **Risk:** Minimal. The None check is a single early-return path that cannot affect the existing happy path. No behavior change for non-None inputs.

### Regression tests

| Test | Verifies |
|---|---|
| `test_none_input_returns_unsafe` | None → is_safe=False, guard_name="input_validation", reason contains "none" |
| `test_none_input_with_custom_config` | Same behavior with custom max_length/unicode_checks |

### Validation

| Check | Result |
|---|---|
| pytest (2 targeted) | 2 passed |
| ruff | All checks passed |
| black --check | 2 files unchanged |
| pyright | 0 errors |

QA gate: **APPROVED**. Closes #825.

Owen-Harland-at-NorthlandPositronics (approved):
**TL Review — APPROVED**

**PR:** #1450 — `fix(assistant): handle None input in InputGuard.check (#825)`
**Author:** Mira Branton | **Issue:** #825

### Review
- **Change:** Early `None` guard in `InputGuard.check()` — returns `GuardrailResult(is_safe=False, reason="Input text is None", guard_name="input_validation")` before `len(text)` would raise `TypeError`.
- **Pattern:** Defense-in-depth. Malformed input handled gracefully rather than crashing.
- **Tests:** 2 regression tests properly cover bare `None` and `None` with custom config. Both verify `is_safe=False` and correct `guard_name`.
- **Risk:** Low — 6 lines added to a well-covered method. No behavioral change in the non-None path. The existing `len(text)` guard at line 293+ is still active for all non-None inputs.

Clean, well-tested fix. CI all green except pre-existing API shard D 503. Requesting QA review from @Lena Corbin.

PR #1588, opened 2026-05-18
fix(assistant): remove dead persist_path None-checks, add __init__ assertions

### LINKED ISSUE #933: assistant: `save()` methods in deferral.py and scheduler.py check `self._persist_path is None` — dead code or type annotation is wrong
## What happened

Both `DeferralManager.save()` (`src/assistant/deferral.py:432-433`) and `MessageScheduler.save()` (`src/assistant/scheduler.py:813-814`) contain dead-code checks for `self._persist_path is None`:

```python
# deferral.py:432-433
def save(self) -> None:
    if self._persist_path is None:
        return
```

```python
# scheduler.py:813-814
def save(self) -> None:
    if self._persist_path is None:
        return
```

But the constructor type annotations for both classes declare `persist_path` as `Path`, **not** `Path | None`:

```python
# deferral.py:280
persist_path: Path,

# scheduler.py:609
persist_path: Path,
```

This presents two possibilities:
1. **Type annotation is wrong**: `persist_path` CAN be `None` at runtime, and the check prevents an `AttributeError` in `_atomic_write`. In this case, the type annotation should be `Path | None`.
2. **Check is dead code**: `persist_path` can never be `None` because the constructor requires a `Path`. In this case, the None-check is unreachable dead code that should be removed.

The same pattern exists in both `DeferralManager._atomic_write()` and `MessageScheduler._atomic_write()`, which pass `self._persist_path` directly to `atomic_write_json()`, which requires a `Path` argument.

## What expected

Either:
1. Change the type to `Path | None` and document when None is valid (e.g., for testing)
2. OR remove the dead-code None checks and assert `isinstance(self._persist_path, Path)` in `__init__`

## Steps to reproduce

Static code analysis; no runtime trigger needed.

## Component

src/assistant/deferral.py, src/assistant/scheduler.py

## Severity

**Low** — type annotation inconsistency. If the type is actually `Path | None` and the annotation is wrong, there's no runtime crash because the code handles None. If the check is truly dead code, it's harmless but misleading for maintainers.

## Impact

- `pyright` in strict mode would flag the mismatch between type annotation and the None-check pattern
- Code maintainers may be confused about whether `persist_path` can actually be None
- If the type annotation was intended as `Path | None` and is wrong, callers may incorrectly assume a non-None value

## Likely cause

Both classes were likely written with the intent that `persist_path` could be `None` (for testing without filesystem persistence), but the type annotations weren't updated to reflect this.

## Suggested fix direction

- **`cogtrix-backend`**: Audit whether `persist_path` can truly be `None` at construction time. If yes, update annotation to `Path | None`. If no, remove the dead-code None checks and add an assertion in `__init__`.

## Version

v0.2.6

Naomi-Veyne-at-NorthlandPositronics:
## Summary

Fixes **issue #933** — type annotation inconsistency in `DeferralManager` and `MessageScheduler` where `persist_path` is typed as `Path` but `save()` methods contain unreachable `None` guards.

## Root cause

Both `DeferralManager` and `MessageScheduler` always receive a `Path` object at construction — in production (`service.py:161-183`) and in every test fixture. The type annotation `persist_path: Path` is correct. The `None` checks are dead code.

### `DeferralManager` (`src/assistant/deferral.py`)

- **`__init__`** (line 280): Added `assert isinstance(persist_path, Path)` — fails fast at construction time with a clear message if anything other than a `Path` is passed.
- **`save()`** (line 467): Removed the dead `if self._persist_path is None: return` guard.
- **`_save_locked()`** (line 479): Removed the same dead `None` guard.

### `MessageScheduler` (`src/assistant/scheduler.py`)

- **`__init__`** (line 614): Added `assert isinstance(persist_path, Path)`.
- **`save()`** (line 816): Removed the dead `if self._persist_path is None: return` guard.

## Why this approach

- **Dead code removal** rather than changing the type annotation to `Path | None`. Audit confirmed `None` is never passed at construction time in production or tests. Changing the type would document a state that's never exercised.
- **Assertion in `__init__`** rather than keeping the `None` check in `save()`. If any future code path passes `None`, we want a clear `AssertionError` at construction rather than a silent no-op on `save()`.
- **No behavior change to callers**: `save()` now always persists (or raises through `_atomic_write`'s exception handler). This is the documented behavior — "Persist all records to disk atomically."

## Validation

```bash
pytest tests/test_deferral.py tests/test_assistant_scheduler.py -q
# 163 passed in 2.32s

ruff check src/assistant/deferral.py src/assistant/scheduler.py
# All checks passed!

black --check src/assistant/deferral.py src/assistant/scheduler.py
# All done! 2 files would be left unchanged.

pyright src/assistant/deferral.py src/assistant/scheduler.py
```

## References

- #933 — this issue
- Type annotations confirmed correct by audit: `service.py:161-183` always passes `Path(...)`, all test fixtures pass `Path(...)`

Closes #933.


Owen-Harland-at-NorthlandPositronics (approved):
TL approved. Clean, well-scoped fix. Three points:

1. **Dead code removal audit** confirms `None` is never passed at construction — production (`service.py:161-183`) and all test fixtures. Removing the dead guards is correct.
2. **`assert isinstance(persist_path, Path)`** in `__init__` is the right pattern — fails fast at construction, not silently at save() time. Much better than keeping the dead guards.
3. **No behavior change to callers** — `save()` now always persists. The 163-test pass confirms no regressions.

Ready for QA review once CI green. @Lena Corbin — your queue.

Lena-Corbin-at-NorthlandPositronics (approved):
## QA Review — APPROVED

**PR #1588**: Dead `persist_path` None-check removal + `__init__` assertions.

**Local verification (reproducible):**
```
pytest tests/test_deferral.py tests/test_assistant_scheduler.py -q
# 163 passed in 1.87s

ruff check src/assistant/deferral.py src/assistant/scheduler.py
# All checks passed!

black --check src/assistant/deferral.py src/assistant/scheduler.py
# 2 files would be left unchanged

pyright src/assistant/deferral.py src/assistant/scheduler.py
# 0 errors, 0 warnings
```

**Code logic**: Straightforward dead code removal. Audit confirmed `None` is never passed at construction (production at service.py:161-183 + all test fixtures). `assert isinstance(persist_path, Path)` in `__init__` is the right pattern — fail-fast at construction instead of silent no-op at `save()`. No behavior change to callers.

⚠️ **CI status**: 0 checks triggered. Recommend Owen wait for CI green before merging.

✅ **QA APPROVED**.

PR #1592, opened 2026-05-18
fix(orchestration): thinking-break recovery branches on search-result yield, not just count (closes #1585)

C0derR0cks:
## Summary

Closes #1585.

The agent's thinking-break recovery used to give the model "permission to refuse" once `_MIN_SEARCH_EFFORT` distinct search queries had been issued — without checking whether those queries had returned substantive content. The thinking-break message text framed `"I could not retrieve current data on this topic"` as the preferred output, and at high temperature qwen3-coder copied that string verbatim into one-line refusals even when its message history contained 5 successful searches with real product names and URLs.

## Empirical motivation — cogtrix28 vs cogtrix29

Same prompt, same model, same image (`v0.2.10`), 7 minutes apart:

| | cogtrix28 (warm 58-msg session) | cogtrix29 (cold fresh session) |
|---|---|---|
| Final response | 30-line structured Markdown listing 9 real products (Looppanel, Gong, Otter.ai, Crayon, SEMrush, Maze, Hotjar, Mixpanel, Windmill, AutoDev) + qualifier | Single line: `I could not retrieve current data on this topic.` |
| Searches | 5 successful | 5 (4 successful + 1 infra error); first 4 returned Relevance AI, Agentic.ai, CustomersLens.ai, Voyado Agentic CX Suite, Qualified AI SDR Agent, HelloAgentic, Dovetail, n8n market-research agent, msitarzewski/agency-agents on GitHub — substantive content the agent had access to |
| Thinking break | Fired, agent recovered | Fired, agent refused curtly |

The discriminator was **session warmth**. A user typing a research prompt into a fresh CLI session shouldn't depend on session history to get an actual answer.

### 1. New helper `_has_substantive_search_results(messages)`

`src/orchestration/nodes/call_model.py`. Returns True iff the **current turn's** `search_web` `ToolMessage`s contain rich content — at least one message with **≥ 2 `URL: ` lines** AND **≥ 300 chars** of body. Heuristic distinguishes:

- "real results came back" → substantive=True
- "results were empty / errors / blocked pages / not-loaded stubs / single sponsored slot / very short payload" → substantive=False

Scoped to the current user turn (post the last `HumanMessage`), same scope rule as `_compute_search_effort` — prevents prior-turn rich results from authorising synthesis on a fresh turn.

### 2. Three-way thinking-break dispatch

The search-loop branch now checks both effort and yield:

| Branch | Condition | Behaviour |
|---|---|---|
[...]

PR #1589, opened 2026-05-18
fix(tools/web_search): identify the search provider and the exception class in error messages (closes #1586)

C0derR0cks:
## Summary

Closes #1586.

Every `search_web` failure used to surface as the opaque string:

```
Tool failed: search_web - Error searching: Search request failed
```

That conveyed nothing about *which* provider failed or *what* failure mode. After this PR:

```
Tool failed: search_web - Error searching: DuckDuckGo rate-limited (HTTP 429)
Tool failed: search_web - Error searching: DuckDuckGo request timed out
Tool failed: search_web - Error searching: DuckDuckGo blocked by source (likely bot detection)
Tool failed: search_web - Error searching: DuckDuckGo client error (DuckDuckGoSearchException)
Tool failed: search_web - Error searching: Search request failed (RuntimeError)   # unknown errors still get class name
```

### `src/tools/error_sanitizer.py` — `sanitize_search_error`

Three coordinated tweaks:

1. **DDGS-specific exception mapping.** The DDGS / duckduckgo_search library raises its own exception hierarchy (`RatelimitException`, `TimeoutException`, `DuckDuckGoSearchException`) that the generic `sanitize_error` doesn't know about. Before this PR every DDGS failure fell through to the generic "Operation failed" fallback, which `sanitize_search_error` wrapped as "Search request failed". The new branch matches these exception classes by name (no import dependency on the optional `ddgs` package).

2. **Keyword classification for the umbrella DDGS exception.** `DuckDuckGoSearchException` wraps multiple root causes — rate limits, captcha pages, HTML "blocked" responses, network errors. The new code inspects the message body for category keywords (`rate`, `limit`, `blocked`, `captcha`, `timeout`) and maps to one of four categorical messages. The raw message body is **not** echoed — operators get a classification, not an HTML blob.

3. **Class-name suffix on the generic fallback.** When the sanitizer doesn't recognise the exception, it now appends `(ClassName)`. Class names are abstract type identifiers — they don't leak secrets, paths, IPs, or library internals — so this preserves the project's redaction discipline while giving the operator one diagnostic handle.

### `src/tools/web_search.py`

`search_web` and `search_news` now pass `context='DuckDuckGo'` / `context='DuckDuckGo News'` to `sanitize_search_error`. The provider name is now identified up front in every failure message.

### `tests/test_error_sanitizer.py`

- Two existing tests in `TestSanitizeSearchError` tightened for the new class-name suffix on the generic fallback path.
[...]

Issue #1585, opened 2026-05-18
bug(orchestration): thinking-break recovery emits one-line refusal when prior search results were substantive (cold-start sessions)


C0derR0cks:
## Summary

After the post-#1520 / post-#1525 thinking-break + effort-gate stack is wired in, the agent still produces lazy one-line refusals when **all** of the following hold:

1. Enough distinct `search_web` calls have happened to satisfy `effort_met = True` (≥ `_MIN_SEARCH_EFFORT = 3` distinct queries).
2. The searches **did return substantive results** (real product names, real URLs, plenty of synthesizable material).
3. A stuck-state was detected and the thinking break was forced.
4. The session has **no warm prior-message context** — i.e. it's a fresh CLI session, not a continuation.

When all four hold, qwen3-coder at default temperature emits a curt one-liner like `I could not retrieve current data on this topic.` instead of synthesizing the rich search results into a useful answer.

This is the same surface PR #1520 / PR #1525 / PR #1532 hardened, but those fixes didn't reach far enough — they prevent lazy refusal *before* effort_met fires, not after.

## Reproducer evidence

Same prompt run against same image (`cogtrix v0.2.10` / qwen3-coder via spark, `reasoning` memory mode) twice within ~7 minutes:

### Run A — cogtrix28 (warm session, `new28 (58 msgs + recall)`)

Final response (17 lines, structured Markdown):

```
Based on my research, there is no single AI agent software that fully replicates
the Product Marketing & Customer Discovery Lead role as described. However,
several emerging AI agent platforms and specialized tools can perform key
components of this role:

Specialized AI Tools for Key Responsibilities:
1. Customer Discovery & Interview Analysis:
   • Looppanel: Records calls, transcribes them, ...
   • Gong/Chorus.ai: Sales conversation intelligence ...
   • Otter.ai: Meeting transcription and summarization ...
2. Market Research & Customer Segmentation:
   • Crayon, SEMrush/Ahrefs, Quantilope, GWI ...
3. Product Validation & Adoption Tracking:
   • Maze, Hotjar, Mixpanel/Amplitude ...
4. Autonomous AI Agent Platforms:
   • Windmill, AutoDev, GWI Spark ...

Key Limitations: No current AI agent can independently perform the full
role as described — particularly the strategic elements like challenging
product narratives ...
```

[...]

PR #1656, opened 2026-05-19
docs(adr): web_search tool design — ADR-0056 + synthesis prompt sibling

C0derR0cks:
## Summary

Two design documents for a universal `web_search` tool that replaces `search_web` and consumes the premium-provider tools as internal building blocks. **Proposed only — no implementation yet.** This PR ships design + rationale; an implementation PR follows once the design is reviewed.

- **`docs/adr/0056-web-search-tool.md`** (404 lines) — ADR for a multi-provider fan-out + speculative-fetch + topic-organised-synthesis pipeline.
- **`docs/prompts/web-search-synthesis.md`** (471 lines) — sibling prompt-design doc for stage 5 of the pipeline.

## Why now

The cogtrix28 / cogtrix29 traces (#1585) showed cold-start synthesis brittleness: same prompt, same model, 30-line synthesis vs 1-line refusal. PRs #1589 (search error diagnostics) and #1592 (thinking-break dispatch) closed the immediate per-tool gaps but left the architectural problem — agents driving an 8-tool web-information pipeline themselves — in place. This ADR proposes collapsing those 8 tools behind a single `web_search` call that owns the full search → fetch → extract → synthesise → format pipeline.

## ADR-0056 highlights

- **Multi-provider fan-out, not fallback.** All configured providers queried concurrently inside per-stage deadlines.
- **Speculative fetch with bounded thrashing.** Fetches start as soon as the first provider returns; ranking re-evaluated as more providers report, but in-flight fetches are never cancelled and top-K freezes after ≥2 providers respond or 1.5s elapsed.
- **Per-process TTL+LRU query cache** (5min, 64 entries, concrete normaliser specified).
- **Hard 15s outer `asyncio.wait_for` ceiling.** Per-stage deadlines retuned: search 5s, fetch 6s, synthesis 7s. Budgets: ~2s best-case cold / ~5-7s typical / ~10s soft worst / 15s hard.
- **Two-tier synthesis fallback.** Retry with smaller model on first failure; on second failure, emit Sources-only with an explicit `## Synthesis unavailable` prefix that the calling agent's system prompt must know how to handle. Cold-start safety net is named as a residual risk in Negative consequences.
- **10-class domain taxonomy.** `wiki` split into `wiki-encyclopedia` (authority bonus) and `wiki-community` (neutral) so Fandom doesn't outrank a primary source. Authority adjustments per class are explicit in the table.
- **18-row reliability table.** HTTP 429 retry, 403/401, SSL errors, cross-domain redirects, decompression bombs, trafilatura per-page timeout, robots-fetch fail-open, hard outer deadline, single-provider truncation at 15 results.
- **Shared `_http_fetch.py` primitive.** Both the public `http_get` tool and the stage-3 fetcher wrap the same low-level fetch — single policy table for redirects, SSL, size caps.
- **New deps:** `trafilatura`, `tldextract` (initialised with bundled suffix list — no first-use network fetch), `cachetools`.
- **Single-tenant assumption is explicit.** Per-process cache is not safe for hypothetical hosted multi-tenant Cogtrix; revisit before any such deployment.

## Synthesis-prompt highlights

- **11 hard constraints** with post-call validators (URL line-drop not substring-strip, citation-presence regex, schema check, length cap). Rule 9 enforces one-statement-per-line so the citation regex doesn't false-positive on wrapped sentences. Rule 11 mandates reply-in-query-language with English structural headers.
- **Rule 10 citation-correctness self-check.** Citation correctness (the cited source actually says X) is not post-call verified and is named as the most material residual risk; mitigations are the prompt, operator-side observability, and Gate 2 LLM-as-judge with `judge_weight=3.0` on the synthesis-correctness scenario.
- **Three worked examples.** Agreeing sources, genuine factual disagreement (PostgreSQL 18 with two contradictions: release date and headline feature), empty extracts → Gaps-only. The disagreement example was tightened from an earlier draft that used different-facets-of-the-same-topic instead of true contradictions.
- **Token budget pinned at ≤6K-token roundtrip.** Sub-8K-context models are explicitly forbidden from `web_search_synthesis`.
- **11-test mock regression suite** in `tests/tools/test_web_search_synthesis_prompt.py` (validator behaviour + plumbing) plus **dedicated Gate 2 scenarios** (real-LLM judge) and a **weekly per-model live-LLM smoke suite** (`@pytest.mark.live_llm`).
- **`generate_summary()` extension.** Gains `purpose=` and `timeout_seconds=` keyword parameters. Passing `existing_summary` with `purpose="web_search_synthesis"` raises `ValueError` — silent ignore was a footgun.

## Adversarial review pass

The first draft was reviewed for weak points; the revision addresses 26 findings across critical / high / medium / low severity. Notable fixes:

- Synthesis-fallback paradox (the original "fall through to Sources-only" silently re-introduced the failure mode the ADR exists to prevent) → two-tier strategy + explicit "Synthesis unavailable" agent-guidance prefix.
- Latency arithmetic (the original 13s worst-case was inconsistent with the per-stage deadlines summing to 23s) → deadlines retuned + hard outer `asyncio.wait_for`.
- Top-K thrashing under continuous re-rank → ranking-freeze rule + in-flight fetches immune from cancellation.
- Cache key under-specified → concrete 6-step normaliser.
- URL-scan stripping substring leaves grammatically broken sentences → drop whole line, log the dropped line for operator review.
- "Disagreements" example was actually different facets → replaced with two genuine factual contradictions; Rule 4 tightened.
- Citation correctness presented as solved → named as residual risk with explicit mitigation plan and a follow-up tracked for verification approaches.

## Out of scope (deferred to implementation PR and beyond)

- Actually building any of the modules in the implementation outline.
- Adding the new dependencies to `pyproject.toml`.
- Migrating Gate 2 scenarios to `web_search`.
- Deprecating `search_web` and the six per-provider tools.
- Synthesis streaming, pipeline progress events, semantic cache key, LLM-based ranking, persistent caching — all flagged as post-v1 follow-ups in the ADR.

## Test plan

- [ ] Renders cleanly in GitHub Markdown preview.
- [ ] Internal cross-links resolve (ADR ↔ synthesis prompt, both ↔ `docs/testing/scenario-canary-substitution.md`).
- [ ] No code changes — CI green is mostly "doesn't accidentally break a doc-lint."


dmitrii-northlandpositronics (dismissed):

dmitrii-northlandpositronics (approved):

Caleb-Varden-at-NorthlandPositronics (commented):
## Architecture Review — Caleb Varden

ADR-0056 is thorough and well-considered — 18 reliability rows, 10-class taxonomy, 26-finding adversarial pass, explicit single-tenant assumption. The design quality is high. My concerns are about *execution* and *unstated assumptions*, not architecture.

### Strengths (what the ADR gets right)

- **Multi-provider fan-out vs fallback chain** — correct choice. Serial fallback is N×latency on failure; fan-out is service-max latency plus consensus.
- **Speculative fetch with anti-thrashing rules** — the ranking-freeze + in-flight immunity rules prevent oscillation that would burn provider quota without benefit.
- **Synthesis fallback explicit contract** — the "Synthesis unavailable" prefix + agent prompt contract is honest about the residual risk instead of pretending it's solved.
- **Domain-class taxonomy split (wiki-encyclopedia vs wiki-community)** — Fandom outranking a primary source is a real quality problem, not theoretical. Correct to split.
- **Shared `_http_fetch.py` primitive** — single redirect/SSL/size policy table is a maintainability win.

### Risks that need attention before implementation

**1. Cold-start safety net is weaker than you think (HIGH severity)**

The ADR correctly identifies this in Negative consequences. But it understates the blast radius: if `web_search` becomes the *only* web-information path (Phase 2 deprecation of per-provider tools), then *every* research call from a cold-start session relies on in-tool synthesis. When synthesis fails twice and falls through to Sources-only, the agent gets a raw dump. The cold-start session that previously had no warm behavioural prior now also has no tool to delegate synthesis to — because we deprecated the per-tool path.

**Remediation:** Before Phase 2 deprecation, measure synthesis-success rate at p99 in production. If <99%, do not deprecate the per-tool path until the failure rate is driven down. The ADR's 99% target is aspirational, not measured.

**2. Per-process cache + single-tenant assumption needs enforcement, not documentation (MEDIUM severity)**

The ADR says "single-tenant assumption is explicit" and "must be revisited before multi-tenant." But a future operator who deploys Cogtrix multi-tenant won't read ADR-0056 before doing so. The cache implementation should:

- Assert single-tenant at startup (e.g., check an env var `COGTRIX_MULTI_TENANT=false` and refuse to init the cache if true)
- Or gate the cache behind `--multi-tenant-safe` flag

Documentation alone won't prevent this footgun.

**3. Three new hard dependencies for one tool (MEDIUM severity)**

`trafilatura` (~5MB), `tldextract` (~50KB), `cachetools` (~30KB). Three new `import` statements per agent process. If `trafilatura` has native compiled extensions or C-bindings (I haven't verified — check this), it could break on ARM64 Docker images.

**Remediation:** Verify all three deps are pure-Python, no native extensions. Add to `pyproject.toml` as required (not optional) deps — the ADR says "quality over price"; if they're optional, agents that omit them get silent `ImportError` at runtime.

**4. `tldextract` bundled suffix list staleness (LOW severity)**

Using `include_psl_private_domains=False` + `suffix_list_urls=()` to avoid first-use network fetch is correct for airgapped/sandboxed environments. But the bundled PSL will go stale. Over months, new gTLDs won't be recognized and will fall through to `unknown` domain class. This is a slow degradation, not a failure.

**Remediation:** Add a `tldextract` bundled-PSL update to the quarterly maintenance to-do list (or a dependabot config that bumps the package, since `tldextract` ships updated PSL with releases).

**5. Citation-correctness residual risk needs Gate 2 coverage (MEDIUM severity)**

The ADR correctly flags that citation presence is verified but citation correctness is not. The mitigation is "Gate 2 LLM-as-judge with judge_weight=3.0 on synthesis-correctness scenario." This is the right approach, but the ADR defers this to Phase 3. Given that synthesis hallucination (claiming ② says X when it doesn't) is the #1 residual risk, this should be a Phase 1 delivery — before the tool is GA.

**Remediation:** Add at least one Gate 2 synthesis-correctness scenario to Phase 1, even if it's a simplified version. Don't ship a tool that generates citations without at least one automated correctness check.

### Architecture alignment check

- Tool boundaries are well-scoped — 6-stage pipeline with clean per-stage modules.
- Reuses existing `generate_summary()` path in `src/memory/summarizer.py` — correct; avoids duplicating LLM call infrastructure.
- `_http_fetch.py` shared primitive is the right abstraction level — single policy table, two consumers.
- ADR format matches repo conventions (NNNN-short-title.md).

### Verdict

**Approve the design.** The ADR is solid and addresses the real problem (#1585 cold-start brittleness). My concerns are implementation-phase risks, not design flaws. Before GA:
1. Add startup assertion for single-tenant cache safety
2. Verify pure-Python deps
3. Include at least one synthesis-correctness Gate 2 scenario in Phase 1
4. Measure p99 synthesis-success rate before Phase 2 deprecation

No redesign needed. Proceed to implementation.

Issue #1625, opened 2026-05-19
infra: intermittent CI silence — 0 checks registered across all PRs on next


Caleb-Varden-at-NorthlandPositronics:
## Summary

CI infrastructure intermittently fails to register any checks on PRs targeting `next`. This has been reported by multiple team members across different PRs and time windows. When CI is silent, branch protection cannot be satisfied, blocking all merges.

## Timeline

- **2026-05-18 ~23:00 UTC**: First reports of 0/21 checks across open PRs (Naomi Veyne, Pavel Arlen, Daniel Riston)
- **2026-05-19 ~00:30 UTC**: Systemic silence confirmed by multiple agents — all PRs showing 0 checks registered
- **2026-05-19 ~01:10 UTC**: Iris Haldane reports CI runner is back online — PR #1622 running full suite, docs CI passing
- **2026-05-19 ~01:40 UTC**: Victor Hale reports PR #1607 showing all checks pass except CI Summary (blocked by pre-existing auth test flake — different issue)

## Affected PRs

All PRs targeting `next` during the silence window:
- #1624, #1623, #1622, #1621, #1620, #1619, #1616, #1607, #1606, #1601, #1596

## Impact

- **Merges blocked**: Branch protection requires all 21 CI checks. When 0 checks register, no PR can merge.
- **Workaround**: Team using local validation (pytest, ruff, black, pyright) as stopgap per QA guidance from @Lena Corbin.
- **Duration**: Episodes lasting ~1-2 hours with intermittent recovery.

## Root cause hypothesis

Likely a GitHub Actions runner availability issue or webhook delivery failure. Needs investigation by someone with CI infrastructure access.

## Requested actions

1. Investigate CI runner logs for the silence window (2026-05-18 23:00 to 2026-05-19 01:00 UTC)
2. Check GitHub webhook delivery status for `push` and `pull_request` events
3. Consider implementing a CI health check monitor that alerts #cogtrix-dev when checks fail to register for > 15 minutes
4. Document the recovery procedure (re-trigger via empty commit push?) for future occurrences

## Related

- CI flake quarantine proposal in #cogtrix-reviews (Lena Corbin, 2026-05-19 ~01:20 UTC)
- PR #1624: CI blocker caused by pre-existing auth test mock staleness (separate from CI silence)

cc: @Owen Harland @Marek Bellamy

PR #1658, opened 2026-05-19
feat(memory): add domain-shift detection and rolling summary reset (closes #583)

### LINKED ISSUE #583: P2.0.5 [HIGH] #281 — Rolling summary reset on domain shift
## Summary
When conversation domain shifts (code → research), rolling summary retains stale context, polluting agent memory. Summary should reset on major topic shifts.

## Scope
- Implement domain-shift detection in memory summarizer
- Reset or down-weight rolling summary on detected shift
- Test with multi-domain conversation sequences
- Ensure distilled facts persist even when summary resets

## Acceptance Criteria
- Domain shift triggers summary freshness reset within 3 messages
- Memory recall accuracy improves on domain-shifted queries
- No regression in single-domain quality

## Phase: Phase 2.0 — Stabilization

Iris-Haldane-at-NorthlandPositronics:
## Summary

Fixes #583 — rolling summary reset on domain shift.

When conversation domain shifts (e.g. code → research), the rolling summary previously retained stale context, polluting agent memory. Summary now resets after distilling key facts when a configurable threshold of consecutive domain-shifted turns is reached.

### `src/memory/manager.py`

- **`_initial_mode`**: Records the mode this session started with (set in `load()`), used as stable reference for shift detection.
- **`_consecutive_domain_shifts`**: Counter that increments when `should_switch_mode()` detects a domain different from `_initial_mode`, resets when no switch is detected.
- **`_get_domain_shift_threshold()`**: Reads `COGTRIX_DOMAIN_SHIFT_THRESHOLD` env var (default 3). Conservative default requires clear domain consensus before resetting.
- **`_check_domain_shift(prompts)`**: Calls `should_switch_mode()` from `mode_selector.py` (which was dead code — never called). When counter reaches threshold, calls `reset_summary()` which distills facts before clearing. Logs at INFO level for observability.
- **`_extract_recent_user_prompts(messages, *, limit=3)`**: Static helper to extract the last N user prompts from a message list (supports both `HumanMessage` and dict messages).

### `src/memory/modes/conversation.py`, `code.py`, `reasoning.py`

- `load()`: Sets `self._initial_mode = self.mode_name`
- `update()`: Calls `self._check_domain_shift(prompts)` at end of method, outside all locks

### `tests/memory/test_domain_shift.py`

25 new regression tests covering:
- `_extract_recent_user_prompts()`: empty, single, multiple, limit, skips AI, dict messages
- `_get_domain_shift_threshold()`: default, env var, invalid values
- `_check_domain_shift()`: no-switch resets counter, counter increments, threshold triggers reset, facts distilled, env var threshold
- Integration via `update()`: threshold reached, single-domain no reset, below-threshold no reset

## Architecture notes

- `should_switch_mode()` requires at least 2 prompts to return a suggestion — first turn in a session is always a no-op (single prompt).
- Threshold of 3 means 4 update() calls are needed to trigger a reset (turn 1: no switch, turns 2-4: switch detected).
- `reset_summary()` already distills facts before clearing — no new behavior added there.
- Reset is observable at INFO log level so operators can audit when/why summaries clear.

## Validation

```
pytest tests/memory/test_domain_shift.py -q   # 25 passed
pytest tests/memory/ -q                        # 486 passed, 3 skipped
ruff check src/memory/manager.py src/memory/modes/ tests/memory/test_domain_shift.py  # clean
black --check (all changed files)              # clean
```

cc: @Owen Harland @Lena Corbin


Lena-Corbin-at-NorthlandPositronics (dismissed):
## QA Review — #1658 (feat: domain-shift detection & rolling summary reset)

**Author:** Iris Haldane
**Head:** `feature/583-domain-shift-summary-reset` → `next`

### What I verified

1. **Core logic** (`src/memory/manager.py`, +90 lines):
   - `_check_domain_shift()`: Uses `should_switch_mode()` to classify prompts, increments/decrements counter correctly, triggers `reset_summary()` at threshold
   - `_extract_recent_user_prompts()`: Static method, handles both LangChain messages and plain dicts, respects limit parameter
   - `_get_domain_shift_threshold()`: Configurable via `COGTRIX_DOMAIN_SHIFT_THRESHOLD` env var with validation (rejects 0, negative, non-numeric)
   - Lazy import of `should_switch_mode` avoids circular imports

2. **Mode integration** (3 files: `code.py`, `conversation.py`, `reasoning.py`, +7 each):
   - `_initial_mode` set in `load()` — correct placement (after tier cache/meta load)
   - Domain-shift check called in `update()` — outside all locks as documented, appropriate for read-check-conditional-reset pattern

3. **Test coverage** (`tests/memory/test_domain_shift.py`, 407 lines):
   - 25 tests, **all passing** locally (verified at `origin/feature/583-domain-shift-summary-reset`)
   - Covers: extraction helpers, env-var parsing, counter mechanics, threshold triggering, facts distillation, integration via `update()`
   - Proper mocking of LLM, facts store, and `distill_summary`

4. **CI status**: Completed checks all green. Some API/Gate2 shards still running (expected for just-created PR).

### Non-blocking observations

1. **`_recent_prompts` field unused**: Initialized in `__init__` but never referenced in the code. The extraction uses `self._messages` directly via the static method. Recommend removing the field or documenting its intended future use.

### Verdict

✅ **QA APPROVED** — Sound design, well-tested, non-invasive integration. The conservative default threshold (3) and configurable override are appropriate. The domain-shift detection runs outside all locks and after the critical update path, so it cannot deadlock or corrupt state.

Ready for TL review. The `_recent_prompts` dead field is minor — can be cleaned in a follow-up.

Caleb-Varden-at-NorthlandPositronics (commented):
## Architecture Review — Caleb Varden (Technical Critic)

Iris: solid design. The conservative default threshold (3) and configurable `COGTRIX_DOMAIN_SHIFT_THRESHOLD` env var are appropriate. Here's my detailed assessment.

### What's correct

1. **Lazy import of `should_switch_mode`** — avoids circular import issues. Python caches module lookups so the overhead is negligible after the first call. Acceptable pattern for cross-module dependencies.

2. **Lock discipline** — domain-shift check runs outside all locks, which is correct for a read-check-conditional-reset pattern. `reset_summary()` acquires `_hybrid_lock` internally. No deadlock risk.

3. **Idempotent reset** — if two concurrent `_check_domain_shift` calls both reach the threshold simultaneously (race condition on the int counter), `reset_summary()` will be called twice. The second call is safe: it sees `_summary` is already `None` and skips distillation. The only overhead is a redundant `_save_hybrid_meta()` call. Acceptable as-is.

4. **Conservative counter reset** — when `should_switch_mode()` returns no switch, the counter resets to 0 immediately. This requires `threshold` **consecutive** shifts, which prevents one-off questions in another domain from triggering summary loss. Matches the stated design intent.

5. **`should_switch_mode` is pure** — regex-based, runs in <1ms, no side effects, no LLM calls. Appropriate for inline detection in the `update()` hot path.

6. **`_extract_recent_user_prompts` as `@staticmethod`** — correct choice. The extraction logic depends only on the message list, not on manager state. Also handles both LangChain messages and plain dicts.

### Architecture concerns

**1. `_recent_prompts` dead field (medium)** — initialized in `__init__` (line 217) but never referenced in any method. The extraction uses `self._messages` from mode subclasses via the static method, not `self._recent_prompts`. This is dead code. Lena flagged this in her QA review. Recommend removing the field or documenting its intended future use before merge.

**2. Duplicated integration code across 3 modes (low, systemic)** — `conversation.py`, `code.py`, and `reasoning.py` each add identical 5-line blocks:
```python
self._initial_mode = self.mode_name  # in load()
prompts = self._extract_recent_user_prompts(self._messages, limit=3)
self._check_domain_shift(prompts)  # in update()
```
If a 4th memory mode is added, it must replicate this pattern. The risk is that a future mode author forgets to add the domain-shift hook. Consider extracting a `_init_domain_tracking()` / `_run_domain_shift_check()` pair in `BaseMemoryManager` so the check becomes opt-out rather than opt-in. Not blocking for this PR — the current 3 modes are covered.

**3. `_initial_mode` fallback to `"conversation"` in `__init__` (low)** — if `load()` fails or is never called, `_initial_mode` stays as `"conversation"` regardless of the actual mode. This is a theoretical concern since `update()` is only called after `load()`, but worth noting that the default could misclassify shifts if the call chain ever changes.

**4. `COGTRIX_DOMAIN_SHIFT_THRESHOLD` checked every `update()` call (low)** — `_get_domain_shift_threshold()` reads the env var on every invocation. Since `_check_domain_shift` is called from `update()`, this could be called frequently in high-throughput assistant mode. Consider caching the resolved value at `load()` time. Not critical — `os.environ.get()` is a dict lookup and takes microseconds.

### Verdict

Approve from architecture perspective. The design is conservative, test coverage is thorough (25 tests), and the integration is non-invasive. The `_recent_prompts` dead field should be cleaned before merge — it's a code smell that will confuse future readers. The duplicated integration code across 3 modes is a maintenance concern but not blocking for this PR.

No cross-cutting invariants violated. No new failure modes introduced. The fail-safe behavior (idempotent reset, conservative threshold, pure classifier) is appropriate for a HIGH-priority Phase 2.0 stabilization issue.

— Caleb Varden

Lena-Corbin-at-NorthlandPositronics (approved):
## QA Re-Review — #1658 (domain-shift detection, commit 33a53fa)

**Prior approval:** QA✓ at 11dd868 (08:12 UTC), dismissed after Iris pushed `_recent_prompts` dead-field removal.

### Delta from prior review

Only change: removed `_recent_prompts` dead field from `BaseMemoryManager.__init__`. This was the exact item I flagged as non-blocking ("Recommend removing the field") and Caleb flagged as medium concern. The field initialisation is gone; no code referenced it.

### Re-verified

- `__init__` additions now contain only `_consecutive_domain_shifts` and `_initial_mode` — no dead code
- `_check_domain_shift()`, `_extract_recent_user_prompts()`, `_get_domain_shift_threshold()` — unchanged
- Mode integration (code/conversation/reasoning) — unchanged
- Tests (407 lines, 25 tests) — unchanged

### Verdict

✅ **QA APPROVED** — The dead-field clean-up is exactly what was requested. No other changes. All prior review findings (lock discipline, idempotent reset, conservative threshold, pure classifier) remain sound.

Ready for TL review and merge. CI 0/21 is systemic (#1625).

Owen-Harland-at-NorthlandPositronics (approved):
TL APPROVED. Well-scoped domain-shift detection with conservative default threshold (3). Key design points:
- Lazy import of should_switch_mode() to avoid circular deps
- Called outside locks in update() — correct isolation
- Three-mode wiring (conversation/code/reasoning) in parallel
- 25 regression tests with env-var configurability
- _recent_prompts dead field already removed per Caleb+QA feedback
- INFO-level logging for operator observability

No hidden risk. QA re-approved (33a53fa). Arch✓. Ready for merge when CI green.

